PR #14, opened 2017-11-09
Bittrex

coadotvn:


http/main.go
```diff
@@ -33,14 +34,18 @@ func main() {
 		reserveAddr,
 	)
 
-	fileSigner := signer.NewFileSigner("/go/src/github.com/KyberNetwork/reserve-data/http/config.json")
+	fileSigner := signer.NewFileSigner("./config.json")
```
tranvictor: to make it consistent to our docker deployment, please keep the "/go/src...", you can symlink your path to match it.

exchange/liqui/signer.go
```diff
@@ -2,5 +2,9 @@ package liqui
 
 type Signer interface {
 	GetLiquiKey() string
+	GetBittrexKey() string
+	GetBinanceKey() string
 	LiquiSign(msg string) string
+	BittrexSign(msg string) string
+	BinanceSign(msg string) string
 }
```
tranvictor: This Signer interface is liqui specific, it shouldn't have bittrex specific in it.

PR #21, opened 2017-11-11
add http runner

coadotvn:


data/fetcher/http_runner/runner.go
```diff
@@ -0,0 +1,35 @@
...
+	echan := make(chan time.Time)
+	bchan := make(chan time.Time)
+	runner := HttpRunner{
+		port,
+		echan,
+		bchan,
+	}
+	NewHttpRunnerServer(
+		runner,
+		fmt.Sprintf(":%v", port),
```
tranvictor: conventionally we should use "%d" instead

data/fetcher/http_runner/server.go
```diff
@@ -0,0 +1,51 @@
...
+	timestamp := time.Now()
+	self.runner.eticker <- timestamp
+	c.JSON(
+		http.StatusOK,
+		gin.H{
+			"success":   true,
+			"timestamp": timestamp,
+		},
+	)
+}
```
tranvictor: `bticker` also needs an API to signal it.

coadotvn:
fix pr reviews: port format and & tick route for bchan

data/fetcher/http_runner/server.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (self *HttpRunnerServer) Run() {
+	self.r.GET("/etick", self.etick)
+	self.r.GET("/btick", self.btick)
+	f, err := os.OpenFile("log.log", os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
+	if err != nil {
+		log.Fatalf("Couldn't open log file: %v", err)
+	}
+	defer f.Close()
+	log.SetOutput(f)
```
tranvictor: Please remove the lines from 46-51, they are not necessary. I will also move the similar lines from http server to main. This log is initialized globally, not http server specifically

data/fetcher/http_runner/server.go
```diff
@@ -0,0 +1,64 @@
...
+
+func NewHttpRunnerServer(runner HttpRunner, host string) *HttpRunnerServer {
+	r := gin.Default()
+	r.Use(sentry.Recovery(raven.DefaultClient, false))
+	return &HttpRunnerServer{
+		runner,
+		host,
+		r,
+	}
+}
```
tranvictor: If we decide not to have `start` and `stop` api for runners, we need to trigger `.Run()` for this runner server in an other go routine.

PR #30, opened 2017-12-23
Support socket for getting order book from binance

halink0803:


tranvictor:
I can only review this PR later tonight, so please work on bittrex also

PR #34, opened 2017-12-27
Add prune outdated data from bolt

halink0803:


data/storage/bolt.go
```diff
@@ -113,6 +115,34 @@ func (self *BoltStorage) CurrentPriceVersion(timepoint uint64) (common.Version,
...
+	if k == nil {
+		err = errors.New(fmt.Sprintf("There no version in %s", Bucket))
+		return err
+	}
+	err = b.Delete([]byte(k))
+	if err != nil {
+		panic(err)
+	}
+	return err
+}
```
tranvictor: Ideally we should prune many snapshots until there are only max no. snapshot in the bucket. Pruning only the oldest one wouldn't work if we decide to reduce the number of snapshot during runtime.

data/storage/bolt.go
```diff
@@ -113,6 +115,34 @@ func (self *BoltStorage) CurrentPriceVersion(timepoint uint64) (common.Version,
...
+	b := tx.Bucket([]byte(Bucket))
+	c := b.Cursor()
+	for k, _ := c.First(); k != nil; k, _ = c.Next() {
+		result++
+	}
+	return result
+}
+
+// PruneOutdatedData Remove first version out of database
+func (self *BoltStorage) PruneOutdatedData(tx *bolt.Tx, Bucket string) error {
```
tranvictor: param shouldn't be in capitalized form. Use `bucket` instead

PR #38, opened 2018-01-02
Add limits precision

halink0803:


common/types.go
```diff
@@ -39,6 +39,32 @@ func TimepointToTime(t uint64) time.Time {
...
+	trading TradingFee
+	funding FundingFee
+}
+
+func InitiateExchangeFee(tradingFee TradingFee, fundingFee FundingFee) ExchangeFees {
+	return ExchangeFees{
+		trading: tradingFee,
+		funding: fundingFee,
+	}
+}
```
tranvictor: This should follow go convention, using `NewExchangeFees` instead of initiate...

common/types.go
```diff
@@ -39,6 +39,32 @@ func TimepointToTime(t uint64) time.Time {
...
+
+func InitiateFundingFee(withdraw map[string]float32, deposit map[string]float32) FundingFee {
+	return FundingFee{
+		withdraw,
+		deposit,
+	}
+}
+
```
tranvictor: this one too

exchange/binance.go
```diff
@@ -3,19 +3,23 @@ package exchange
 import (
 	"errors"
 	"fmt"
+	"log"
 	"strconv"
 	"strings"
 	"sync"
 
+	"math/big"
```
tranvictor: Lets group all builtin packages into one group by removing the blank line in the middle

exchange/binance.go
```diff
@@ -37,6 +41,56 @@ func (self *Binance) UpdateDepositAddress(token common.Token, address string) {
...
+				if filter.FilterType == "PRICE_FILTER" {
+					// update price min
+					minPrice, _ := strconv.ParseFloat(filter.MinPrice, 32)
+					exchangePrecisionLimit.PriceLimit.Min = float32(minPrice)
+					// update price max
+					maxPrice, _ := strconv.ParseFloat(filter.MaxPrice, 32)
+					exchangePrecisionLimit.PriceLimit.Max = float32(maxPrice)
+				}
+			}
+			self.exchangeInfo[pair.PairID()] = exchangePrecisionLimit
```
tranvictor: Need to be careful at this map assignment. In the future, `exchangeInfo` might be read during `deposit`, `withdraw`, `trade` action which happen concurrently with this update action. Go map will panic when there are  reads/writes happen concurrently.

Suggestion: introduce a `mutex` here.

exchange/bittrex.go
```diff
@@ -42,6 +49,35 @@ func (self *Bittrex) UpdateDepositAddress(token common.Token, address string) {
...
+	for _, symbol := range symbols {
+		symbolName := symbol.Base + symbol.Quote
+		if symbolName == pairName {
+			exchangePrecisionLimit := common.ExchangePrecisionLimit{}
+			//update precision
+			exchangePrecisionLimit.Precision.Amount = 8
+			exchangePrecisionLimit.Precision.Price = 8
+			// update limit
+			exchangePrecisionLimit.AmountLimit.Min = symbol.MinAmount
+			self.exchangeInfo[pair.PairID()] = exchangePrecisionLimit
```
tranvictor: this one too

PR #50, opened 2018-01-16
Update deposit addresses

halink0803:


tranvictor:
We need to parallelize the deposit update process even though it happens only at the beginning of a run. Imagine we have 20 or even 50 tokens.

halink0803:
@tranvictor I've added concurrent when update deposit addresses. Please review.

exchange/binance.go
```diff
@@ -38,7 +38,8 @@ func (self *Binance) UpdateAllDepositAddresses(address string) {
 	}
 }
 
-func (self *Binance) UpdateDepositAddress(token common.Token, address string) {
+func (self *Binance) UpdateDepositAddress(token common.Token, address string, wait *sync.WaitGroup) {
```
tranvictor: We should not add wait group param into this function. It is not relevant to Binance scope. It is also the same to other exchanges.

cmd/exchanges.go
```diff
@@ -128,16 +153,22 @@ func NewMainnetExchangePool(addressConfig common.AddressConfig, signer *signer.F
 		switch exparam {
 		case "bittrex":
 			bit := exchange.NewBittrex(bittrex.NewRealBittrexEndpoint(signer), bittrexStorage)
+			wait := sync.WaitGroup{}
 			for tokenID, addr := range addressConfig.Exchanges["bittrex"] {
-				bit.UpdateDepositAddress(common.MustGetToken(tokenID), addr)
+				wait.Add(1)
+				go bit.UpdateDepositAddress(common.MustGetToken(tokenID), addr, &wait)
```
tranvictor: Shouldn't pass `&wait` to this function, better have another function and signal `Done` inside of that function. Like this:
```
func AsyncUpdateDepositAddress(..., wait) {
   bit.UpdateDepositAddress(common.MustGetToken(tokenID), addr)
   wait.Done()
}
```
and run it in a go routine:
```
go AsyncUpdateDepositAddress(common.MustGetToken(tokenID), addr, &wait)
```

PR #51, opened 2018-01-16
handle-timestamp

NNagato:


blockchain/blockchain.go
```diff
@@ -265,6 +265,10 @@ func toBlockNumArg(number *big.Int) string {
 	return hexutil.EncodeBig(number)
 }
 
+func uint64ToBig(number uint64) *big.Int {
+	return big.NewInt(int64(number))
+}
```
tranvictor: This function is too small, no need to split it into a function.

blockchain/blocktime.go
```diff
@@ -2,9 +2,16 @@ package blockchain
...
+	"context"
 )
 
-func InterpretTimestamp(blockno uint64, txindex uint) uint64 {
-	// TODO: Using real timestamp
-	return uint64(time.Second)*15*blockno + uint64(txindex)
+func (self *Blockchain) InterpretTimestamp(blockno uint64, txindex uint) uint64 {
+	context := context.Background()
+	block, err := self.client.HeaderByNumber(context, uint64ToBig(blockno))
+	if(err != nil) {
```
tranvictor: it should be `if err != nil`, no need for `()` here. Also there is always a space after a keyword like `if`.

PR #62, opened 2018-01-26
gas price from gas station

NNagato:


blockchain/gas_oracle.go
```diff
@@ -0,0 +1,71 @@
...
+		for {
+			select {
+			case <- ticker.C:
+				self.RunGasPricing()
+			}
+		}
+	}()
+}
+
+func (self *GasOracle) RunGasPricing() {
```
halink0803: This function should return error, log is not enough. Example:
log.Printf("Cannot get GasOracle from Gasstation: %s", err.Error())
return err

PR #76, opened 2018-02-06
Track block number

NNagato:


data/fetcher/fetcher.go
```diff
@@ -292,16 +295,19 @@ func (self *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord)
...
 					activity.Result["tx"].(string),
+					blockNum,
 					"mined",
 					err,
 				}
 			case "failed":
 				result[activity.ID] = common.ActivityStatus{
 					activity.ExchangeStatus,
 					activity.Result["tx"].(string),
+					blockNum,
```
tranvictor: In this scope, you didn't get `blockNum` from anywhere?

data/fetcher/fetcher.go
```diff
@@ -396,6 +403,10 @@ func (self *Fetcher) PersistSnapshot(
 		log.Printf("Aggregate statuses, final activity: %+v", activity)
 		if activity.IsPending() {
 			pendingActivities = append(pendingActivities, activity)
+		} else {
+			if activity.Action == "set_rates" {
+				activity.Result["blockNum"] = activityStatus.BlockNum
+			}
```
tranvictor: Why do you only update block number when action is set_rates? You can just update the block number for all of the txs and just update it.

PR #85, opened 2018-02-09
Remove duplicate get all prices

cyberliem:
Reduced the amount of data return on get all rates by grouping them from Block X to Block Y, provided the following condition:

- Every block between X and Y contains all the same Crypto ID
- With each Crypto ID, the BaseBuy, CompactBuy, BaseSell, CompactSell, Rate are the same for everyt block between X and Y.

Increased the Get_all_rates period form 1 hour to 24 hours.
Clean the code and increase its readability.

### DIFF
--- a/data/reserve_data.go
+++ b/data/reserve_data.go
@@ -85,29 +85,29 @@ func (self ReserveData) CurrentRateVersion(timepoint uint64) (common.Version, er
 	return self.storage.CurrentRateVersion(timepoint)
 }
 
-func compareData(old, new map[string]common.RateResponse) bool {
-	for tokenID, oldElem := range old {
-		newelem, ok := new[tokenID]
+func isDuplicated(oldData, newData map[string]common.RateResponse) bool {
+	for tokenID, oldElem := range oldData {
+		newelem, ok := newData[tokenID]
 		if !ok {
-			return (false)
+			return false
 		}
 		if oldElem.BaseBuy != newelem.BaseBuy {
-			return (false)
+			return false
 		}
 		if oldElem.CompactBuy != newelem.CompactBuy {
-			return (false)
+			return false
 		}
 		if oldElem.BaseSell != newelem.BaseSell {
-			return (false)
+			return false
 		}
 		if oldElem.CompactSell != newelem.CompactSell {
-			return (false)
+			return false
 		}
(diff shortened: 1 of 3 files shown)


data/reserve_data.go
```diff
@@ -85,39 +85,79 @@ func (self ReserveData) CurrentRateVersion(timepoint uint64) (common.Version, er
 	return self.storage.CurrentRateVersion(timepoint)
 }
 
+func compareData(old, new map[string]common.RateResponse) bool {
+	for tokenID, oldElem := range old {
+		newelem, ok := new[tokenID]
```
tranvictor: we should avoid using `new` as a variable name. People might get confused reading it because `new` is actually a keyword

data/reserve_data.go
```diff
@@ -85,39 +85,79 @@ func (self ReserveData) CurrentRateVersion(timepoint uint64) (common.Version, er
...
+			return (false)
+		}
+		if oldElem.CompactSell != newelem.CompactSell {
+			return (false)
+		}
+		if oldElem.Rate != newelem.Rate {
+			return (false)
+		}
+	}
+	return (true)
```
tranvictor: No need to wrap the return value with `()`

data/reserve_data.go
```diff
@@ -85,39 +85,79 @@ func (self ReserveData) CurrentRateVersion(timepoint uint64) (common.Version, er
 	return self.storage.CurrentRateVersion(timepoint)
 }
 
+func compareData(old, new map[string]common.RateResponse) bool {
```
tranvictor: we can change the function name to `isDuplicated` or `isTheSameData` or something like that to increase readability.

data/reserve_data.go
```diff
@@ -85,39 +85,79 @@ func (self ReserveData) CurrentRateVersion(timepoint uint64) (common.Version, er
...
-		}
-		one.Data = data
+		one.Data = getOneRateData(rate)
 		one.BlockNumber = rate.BlockNumber
-		result = append(result, one)
+		//if one is the same as current unchanged block
+		if compareData(one.Data, current.Data) {
+			result[len(result)-1].ToBlockNumber = one.BlockNumber
+
+		} else {
```
tranvictor: remove extra blank line

PR #99, opened 2018-02-22
Add rotate log

halink0803:
- Rotate log using Lumberjack
- 24h hour log file

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"runtime"
+	"time"
 
 	"github.com/KyberNetwork/reserve-data/blockchain"
 	"github.com/KyberNetwork/reserve-data/blockchain/nonce"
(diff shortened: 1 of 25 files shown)


cmd/main.go
```diff
@@ -34,10 +36,34 @@ func loadTimestamp(path string) []uint64 {
...
+	log.SetFlags(log.LstdFlags | log.Lmicroseconds | log.Lshortfile)
+	log.SetOutput(mw)
+
+	rotateTicker := time.NewTicker(24 * time.Hour)
+	go func() {
+		for {
+			<-rotateTicker.C
+			logger.Rotate()
+		}
+	}()
```
tranvictor: So this will only rotate after 24 hours? This is not good, we want to rotate it by day, i mean to a log with date in the file name so we can look into log of a specific date with ease.

halink0803: I've updated it using daily cron to rotate log.

PR #101, opened 2018-02-22
Add time limit 24 hours to get activities

halink0803:


http/server.go
```diff
@@ -648,6 +648,9 @@ func (self *HTTPServer) GetActivities(c *gin.Context) {
 	}
 	fromTime, _ := strconv.ParseUint(c.Query("fromTime"), 10, 64)
 	toTime, _ := strconv.ParseUint(c.Query("toTime"), 10, 64)
+	if toTime == 0 {
+		toTime = MAX_TIMESPOT
+	}
```
tranvictor: This doesn't work, if you set toTime to MAX_TIMESPOT, the api will always exceed max time range. Should set toTime to current time point.

halink0803: Yes, I updated it using current time point.

PR #129, opened 2018-03-13
Avoid stat server to initiate unnecessary instances

NNagato:


cmd/configuration/getconfig.go
```diff
@@ -132,46 +101,106 @@ func GetConfig(kyberENV string, authEnbl bool, endpointOW string) *Config {
...
+			EthereumEndpoint:        statConfig.EthereumEndpoint,
+			BackupEthereumEndpoints: statConfig.BackupEthereumEndpoints,
+			SupportedTokens:         statConfig.SupportedTokens,
+			WrapperAddress:          statConfig.WrapperAddress,
+			PricingAddress:          statConfig.PricingAddress,
+			ReserveAddress:          reserveAddr,
+			FeeBurnerAddress:        burnerAddr,
+			NetworkAddress:          networkAddr,
+			WhitelistAddress:        whitelistAddr,
+			ThirdPartyReserves:      thirdpartyReserves,
```
halink0803: This config belong to stat config.

cmd/clicmd/startserver.go
```diff
@@ -152,9 +154,13 @@ func serverStart(cmd *cobra.Command, args []string) {
 		}
 	}
 
-	//nonceCorpus := nonce.NewAutoIncreasing(infura, fileSigner)
-	nonceCorpus := nonce.NewTimeWindow(infura, config.BlockchainSigner)
-	nonceDeposit := nonce.NewTimeWindow(infura, config.DepositSigner)
+	var nonceCorpus *nonce.TimeWindow
+	var nonceDeposit *nonce.TimeWindow
+	
+	if !(noCore && enableStat) {
```
halink0803: Check again this logic, it should be only !noCore, because nonceCorpus and nonceDeposit does not depend on stat.

PR #133, opened 2018-03-16
Change deposit withdraw interface

halink0803:


exchange/huobi.go
```diff
@@ -363,8 +362,9 @@ func (self *Huobi) DepositStatus(id common.ActivityID, timepoint uint64) (string
 	return "", errors.New("Deposit doesn't exist. This shouldn't happen unless tx returned from huobi and activity ID are not consistently designed")
```
tranvictor: we should also remove this error. It only happens when we embed extra logic (info) into activity ID. WE dont do so now so this error is invalid.

PR #136, opened 2018-03-20
Exchange failure tolerant

halink0803:


data/reserve_data.go
```diff
@@ -192,6 +192,25 @@ func (self ReserveData) GetRate(timepoint uint64) (common.AllRateResponse, error
 	}
 }
 
+func (self ReserveData) GetExchangeStatus() (common.ExchangesStatus, error) {
+	var data common.ExchangesStatus
+	var err error
+	data, err = self.storage.GetExchangeStatus()
```
tranvictor: dont have to declare the vars beforehand without any additional benefits like this.

PR #139, opened 2018-03-26
Walletstats

cyberliem:


http/server.go
```diff
@@ -1585,7 +1589,18 @@ func (self *HTTPServer) GetUserVolume(c *gin.Context) {
 func (self *HTTPServer) GetTradeSummary(c *gin.Context) {
 	fromTime, _ := strconv.ParseUint(c.Query("fromTime"), 10, 64)
 	toTime, _ := strconv.ParseUint(c.Query("toTime"), 10, 64)
-	data, err := self.stat.GetTradeSummary(fromTime, toTime)
+	tzparam, _ := strconv.ParseInt(c.Query("timeZone"), 10, 64)
```
halink0803: fromTime and toTime are not set required and toTime is not set a default value, there will not be an error if user won't provide toTime, it should throw one.

stat/fetcher.go
```diff
@@ -340,13 +346,23 @@ func (self *Fetcher) FetchLogs(fromBlock uint64, toBlock uint64, timepoint uint6
 	}
 }
 
+func (self *Fetcher) updateTimeZoneBuckets(timestamp uint64, updates common.TradeStats) (err error) {
```
halink0803: function return type and one value return bracket is not necessary.

stat/storage/stat_bolt.go
```diff
@@ -172,7 +187,18 @@ func getTimestampByFreq(t uint64, freq string) (result []byte) {
 		result = uint64ToBytes(t / uint64(time.Hour) * uint64(time.Hour))
 	case "d", "D":
 		result = uint64ToBytes(t / uint64(time.Hour*24) * uint64(time.Hour*24))
+	default:
+		// utc timezone
+
+		offset, _ := strconv.ParseInt(strings.TrimPrefix(freq, "utc"), 10, 64)
```
halink0803: Why not parse offset uint then we do not have to double parsing (int64 then uint64) later?

stat/storage/stat_bolt.go
```diff
@@ -256,76 +280,174 @@ func (self *BoltStatStorage) getTradeStats(fromTime, toTime uint64, freq string)
...
+					return err
+				}
+			}
+		}
 		return nil
 	})
 	return err
 }
 
+func findField(fieldName string, stat common.TradeStats) float64 {
```
halink0803: Is it really necesary to separate this into a function?

stat/storage/stat_bolt.go
```diff
@@ -256,76 +280,174 @@ func (self *BoltStatStorage) getTradeStats(fromTime, toTime uint64, freq string)
...
+			avgUsd = float64(stat[strings.Join([]string{"wallet_usd_volume", walletAddr}, "_")]) / tradeCount
+		}
+		result[timestamp] = map[string]float64{
+			"total_eth_volume":     stat[strings.Join([]string{"wallet_eth_volume", walletAddr}, "_")],
+			"total_usd_amount":     stat[strings.Join([]string{"wallet_usd_volume", walletAddr}, "_")],
+			"total_burn_fee":       stat[strings.Join([]string{"wallet_burn_fee", walletAddr}, "_")],
+			"unique_addresses":     stat[strings.Join([]string{"wallet_first_trade_in_day", walletAddr}, "_")],
+			"new_unique_addresses": stat[strings.Join([]string{"wallet_first_trade_ever", walletAddr}, "_")],
+			"kyced_addresses":      stat[strings.Join([]string{"wallet_kyced_in_day", walletAddr}, "_")],
+			"total_trade":          tradeCount,
```
halink0803: I think this should be used fmt.Sprintf("wallet_eth_volume_%s", walletAddr) instead of strings join, save more space and don't have to create an slice of 2 strings every times.

stat/storage/stat_bolt.go
```diff
@@ -382,10 +504,11 @@ func (self *BoltStatStorage) GetUserVolume(fromTime uint64, toTime uint64, freq
...
+func (self *BoltStatStorage) GetTradeSummary(fromTime uint64, toTime uint64, timezone int64) (common.StatTicks, error) {
+	log.Printf("fromtime %v toTime %v fre %v", fromTime, toTime, timezone)
 	result := common.StatTicks{}
-
-	stats, err := self.getTradeStats(fromTime, toTime, "D")
+	tzstring := fmt.Sprintf("%s%d", TIMEZONE_BUCKET_PREFIX, timezone)
+	stats, err := self.getTradeStats(fromTime, toTime, tzstring)
 	if err != nil {
 		return result, err
 	}
```
halink0803: Need to add usage instruction to Readme.md for new api.

stat/storage/stat_bolt.go
```diff
@@ -160,6 +173,8 @@ func getBucketNameByFreq(freq string) (bucketName string, err error) {
 		bucketName = HOUR_BUCKET
 	case "d", "D":
 		bucketName = DAY_BUCKET
+	default:
+		bucketName = freq
 	}
 	return
 }
```
halink0803: This function does not return any error?

http/server.go
cyberliem: This is validated in reserve_data, with validateTimeWindow

stat/fetcher.go
cyberliem: This is variable return, without bracket it will raise syntax error.

stat/storage/stat_bolt.go
cyberliem: Offset can be negative, which is why it must be parsed to int, not uint ( UTC-12 for example).

stat/storage/stat_bolt.go
cyberliem: There is this benchmark test about string join vs fmt.Sprint. Result shows that strings.Join is better in this case. 
https://gist.github.com/dtjm/c6ebc86abe7515c988ec

stat/storage/stat_bolt.go
cyberliem: As it is, no error is return.

http/server.go
halink0803: 1. Function validateTimeWindow in reserve_stat, not reserve_data
2. Please check again logic in validateTimeWindow, you pass freq "D" and it check to limit, not the requirement of toTime value. It might throw "Invalid frequencies" and it is also not the correct error message in this situation.

stat/fetcher.go
halink0803: This one is also a valid function definition, but to keep project consistent about convention, you should use "func funcName(params type) type {
  *err define inside the function body*
}
However this one is not a must change but should (I think)

stat/storage/stat_bolt.go
halink0803: ok, I got it.

stat/storage/stat_bolt.go
halink0803: As you choose base on benchmark, why not choose string concat then?

stat/storage/stat_bolt.go
halink0803: Then why function definition return (err error) as it is always nil?

stat/storage/stat_bolt.go
cyberliem: the syntax was based on previous version from dev branch. I thought I'd keep it unified.

stat/storage/stat_bolt.go
cyberliem: added

PR #147, opened 2018-04-03
Failure resistance

halink0803:
- [x] API for the analytic to push notification. This API has following params: exchange, action, token pair, from, to, isWarning, msg. Exchange is the name of an exchange that analytic wants to notify about. Msg is the message for it. from, to is timestamp of that the notification covers, they are in milisec.

- [x] Dont allow analytic to disable/enable exchange anymore

- [x] API to get all of current notification that analytic has pushed (including all exchanges)

### DIFF
--- a/data/storage/bolt.go
+++ b/data/storage/bolt.go
@@ -1031,32 +1031,37 @@ func (self *BoltStorage) UpdateExchangeStatus(data common.ExchangesStatus) error
 }
 
 func (self *BoltStorage) UpdateExchangeNotification(
-	exchange, action, tokenPair string, fromTime, toTime uint64, isWarning bool, msg string) error {
+	exchange, action, token string, fromTime, toTime uint64, isWarning bool, msg string) error {
 	var err error
-	var notifications common.ExchangeNotifications
+	notifications := common.ExchangeNotifications{}
 	self.db.Update(func(tx *bolt.Tx) error {
 		b := tx.Bucket([]byte(EXCHANGE_NOTIFICATIONS))
 		c := b.Cursor()
 		k, v := c.Last()
-		if v == nil {
-			err = errors.New("Notifications not found")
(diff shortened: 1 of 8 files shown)


halink0803:
[support new failure resistance features](https://trello.com/c/mF54P4WZ/193-support-new-failure-resistance-features)

data/storage/bolt.go
```diff
@@ -1027,3 +1029,61 @@ func (self *BoltStorage) UpdateExchangeStatus(data common.ExchangesStatus) error
 	})
 	return err
 }
+
+func (self *BoltStorage) UpdateExchangeNotification(
+	exchange, action, token string, fromTime, toTime uint64, isWarning bool, msg string) error {
```
tranvictor: lets not encode all notifications into a json string anymore. we should have different bucket for different exchange, each bucket has different key for each pair.

data/storage/bolt.go
```diff
@@ -1027,3 +1029,61 @@ func (self *BoltStorage) UpdateExchangeStatus(data common.ExchangesStatus) error
...
+			IsWarning: isWarning,
+			Message:   msg,
+		}
+		exchangeAction[action] = exchangeToken
+		notifications[exchange] = exchangeAction
+
+		// remove old data
+		if k != nil {
+			b.Delete(k)
+		}
```
tranvictor: Also Delete here is redundant. Put can also overwrite data as i remember.

halink0803: As I chose the key by timepoint then different timepoint will have different key, then it needs to be deleted. However, I will change as you recommended above using exchange for bucket name and action_token for key then we don't need to delete anymore.

data/storage/bolt.go
halink0803: updated

PR #150, opened 2018-04-04
[WIP] Quick code review

favadi:


stat/storage/log_bolt.go
```diff
@@ -128,120 +140,104 @@ func (self *BoltLogStorage) UpdateLogBlock(block uint64, timepoint uint64) error
...
-				return err
+			if vErr := json.Unmarshal(v, &record); vErr != nil {
+				return vErr
 			}
 			result = append(result, record)
 		}
 		return nil
 	})
-	return result, err
+	return
```
tranvictor: Personally I hesitate using named return, it is quite easy to make a bug if the same name var is declared and inited in a different scope.
For example:
```
func A() (err error) {
  if (...) {
    err := some_func_call()
  }
  return
}
```
The compiler will not help us prevent such case.

On the other hand, if we dont use named return like this:
```
func A() (error) {
  if (...) {
    err := some_func_call()
  }
  return err // compiler will help us here
}
```

Personally, i think named return should only be used in a function that has very few vars and very straight forward outcome (very few scopes or branches)

tranvictor:
I just left a few comments, overall, the PR is good.

data/fetcher/fetcher_test.go
```diff
@@ -103,8 +103,8 @@ func TestUnchangedFunc(t *testing.T) {
 }
 
 func TestExchangeDown(t *testing.T) {
-
 	// mock fetcher
+	// TODO: no test_fetcher.db found in repo, where is it?
 	testFetcherStoragePath := "/go/src/github.com/KyberNetwork/reserve-data/data/fetcher/test_fetcher.db"
```
tranvictor: ah I think the reason that travisCI couldn't create that file is because you put the repo in wrong directory. There are a few prerequisites:
1. GOPATH should be `/go`
2. The repo should be at `/go/src/github.com/KyberNetwork/reserve-data`

stat/storage/log_bolt.go
favadi: > i think named return should only be used in a function that has very few vars and very straight forward outcome 

I think the same, and I think the use of named return here is quite appropriate. As you said, named return has its both problems and merits, so it will come to trade-off and personal preferences. I'll remove the use of named return here.

data/fetcher/fetcher_test.go
favadi: `GOPATH should be /go` I don't see any reason to require that, user should have the choice of putting GOPATH everywhere he wants, and generally speaking, hard code directory path like that is an anti-pattern. I'll try to fix this.

favadi: I fixed the problem while keeping same logic with: https://github.com/KyberNetwork/reserve-data/pull/150/commits/3c72198ef11f6f81fb052e23a9c0321f06ab44a1.

But the proper location for temp database in test should not be OS's tmp dir, I did that here: https://github.com/KyberNetwork/reserve-data/pull/150/commits/3c72198ef11f6f81fb052e23a9c0321f06ab44a1

stat/storage/log_bolt.go
favadi: I did the samething here: https://github.com/KyberNetwork/reserve-data/pull/150/commits/e5e99f4db2ea4d54bb44a77dc8c23d2644baf3dc. Tests are now all passed.

stat/util/ip_locator.go
```diff
@@ -1,42 +1,55 @@
...
 
-func init() {
-	var err error
-	IPLocator, err = geoip2.Open(GEO_DATABASE)
+// newIPLocator returns an instance of ipLocator.
+func newIPLocator() (*ipLocator, error) {
+	dbPath := path.Join(common.CurrentDir(), "GeoLite2-Country.mmdb")
+	r, err := geoip2.Open(dbPath)
 	if err != nil {
-		panic(err)
```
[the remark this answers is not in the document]
favadi: That's true, because we embed the maxmind's DB in source tree, this error is a programmer error and should be panic.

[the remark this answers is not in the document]
favadi: This is fixed with https://github.com/KyberNetwork/reserve-data/pull/150/commits/cc154161892f170f5bbdfeabe5491d2c47ff136a.

stat/storage/log_bolt.go
favadi: I reverted the changes to named return: https://github.com/KyberNetwork/reserve-data/pull/150/commits/a3311a042b12499aaba26a7928522d94899b0891.

PR #155, opened 2018-04-09
add err in activities API

cyberliem:


cyberliem:
[update withdraw/deposit activity error](https://trello.com/c/u7khvKv3/199-update-withdraw-deposit-activity-error)

PR #161, opened 2018-04-11
change core activity error return to string

cyberliem:


core/reserve_core.go
```diff
@@ -72,6 +72,10 @@ func (self ReserveCore) Trade(
 		}
 	}
 	uid := timebasedID(id)
+	var errstr string
+	if err != nil {
+		errstr = err.Error()
+	}
```
tranvictor: I have seen quite a few times we use this logic to turn an error into string, could you please make up an util function in `common` package and use it instead?

PR #165, opened 2018-04-12
Binance batch querry

cyberliem:
fix the skipping pair logic err


cyberliem:
[Split requests to exchanges into batches to avoid breaking request limit](https://trello.com/c/9unpPGKk/205-split-requests-to-exchanges-into-batches-to-avoid-breaking-request-limit)

PR #173, opened 2018-04-17
Atomic last block update at fetchlog

cyberliem:


cyberliem:
[also update last block + index atomically on log syncing](https://trello.com/c/260xrz9p/211-also-update-last-block-index-atomically-on-log-syncing)

PR #184, opened 2018-04-20
Stable token params api

cyberliem:
add Stable token params API

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -2332,6 +2332,153 @@ func (self *HTTPServer) GetReserveVolume(c *gin.Context) {
 	)
 }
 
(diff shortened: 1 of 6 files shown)


http/server.go
```diff
@@ -2332,6 +2332,153 @@ func (self *HTTPServer) GetReserveVolume(c *gin.Context) {
...
+		c.JSON(
+			http.StatusOK,
+			gin.H{
+				"success": false,
+				"reason":  "the data size must be less than 1 MB",
+			},
+		)
+		return
+	}
+	err := self.stat.ConfirmStableTokenParams(value)
```
tranvictor: oh no, stable token params should be handle within core scope. Not stat.

PR #187, opened 2018-04-21
Add analytic_dev mode

ductm54:


cmd/configuration/getconfig.go
```diff
@@ -29,7 +29,7 @@ func GetChainType(kyberENV string) string {
 		return "homestead"
 	case "staging":
 		return "byzantium"
-	case "simulation":
+	case "simulation", "analytic_dev":
```
tranvictor: analytic_dev is using ropsten, it is different to simulation which use dev chain. need to handle this differently.

ductm54: analytic_dev using same dev chain as simulation

PR #207, opened 2018-05-02
pricing fee setrate

NNagato:


stat/storage/fee_setrate_bolt.go
```diff
@@ -0,0 +1,208 @@
...
+	return feeSetRate, err
+}
+
+func gasInGWei(st common.StoreTransaction) float64 {
+	wei := float64(st.GasPrice * st.GasUsed) / float64(GWEI)
+	return wei
+}
+
+func getMinMaxTimeStamp(fromTime uint64, toTime uint64) (uint64, uint64) {
+	ftNow := now.New(time.Unix(int64(fromTime), 0).UTC()).BeginningOfDay().Unix()
```
halink0803: This one can easily calculate, no need to use external library like this. Take a look at stat_bolt.go, function getTimestampByFreq

stat/fetcher.go
```diff
@@ -86,10 +91,102 @@ func (self *Fetcher) Run() error {
...
+}
+
+func (self *Fetcher) GetToBlock() uint64 {
+	currentBlock := self.currentBlock
+	if currentBlock == 0 {
+		return 0
+	}
+	if currentBlock - blockNumMarker <= BLOCK_RANGE {
+		blockNumMarker = currentBlock
+		sleepTime = 2 * time.Minute
```
halink0803: This variable set a constant value, why put it in this condition?

stat/fetcher.go
```diff
@@ -86,10 +91,102 @@ func (self *Fetcher) Run() error {
...
+	}
+	toBlock := blockNumMarker + BLOCK_RANGE
+	return toBlock
+}
+
+func makeApi(fromBlock, toBlock uint64) string {
+	fromBlockStr := strconv.FormatUint(fromBlock, 10)
+	toBlockStr   := strconv.FormatUint(toBlock, 10)
+	// log.Println("toBlock: ", toBlockStr)
+	return "http://api.etherscan.io/api?module=account&action=txlist&address=0x8bc3da587def887b5c822105729ee1d6af05a5ca&startblock=" + fromBlockStr + "&endblock=" + toBlockStr +  "&sort=desc&apikey=6AIQWS595MIZN63BXFM1T1XIRV8R7F7A6E"
```
halink0803: What address is this? Put this address and api key to config and use fmt.Sprintf() to make url string. No need to create a function for just a simple expression like this.

stat/storage/fee_setrate_bolt.go
```diff
@@ -0,0 +1,208 @@
...
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	storage := &BoltFeeSetRateStorage{db}
+	return storage, err
+}
+
+func stringToUint64(blockString string) (uint64, error) {
```
halink0803: This is a basic function, no need to separate to a function like this.

stat/storage/fee_setrate_bolt.go
```diff
@@ -0,0 +1,208 @@
...
+				return err
+			}
+			log.Println("tx: ", txs.BlockNumber)
+			latestBlockChecked, err = stringToUint64(txs.BlockNumber)
+			if err != nil {
+				log.Println(err)
+				return err
+			}
+		}
+		return nil
```
halink0803: Return err here

stat/storage/fee_setrate_bolt.go
```diff
@@ -0,0 +1,208 @@
...
+		log.Println(err)
+		return 0, err
+	}
+	return blockNum, nil
+}
+
+func (self *BoltFeeSetRateStorage) GetLastBlockChecked() (uint64, error) {
+	var latestBlockChecked uint64
+	var err error
+	self.db.Update(func(tx *bolt.Tx) error {
```
halink0803: This should be err :=

stat/fetcher.go
```diff
@@ -86,10 +111,97 @@ func (self *Fetcher) Run() error {
...
+	}
+}
+
+func (self *Fetcher) GetToBlock() uint64 {
+	currentBlock := self.currentBlock
+	blockNumMarker := self.blockNumMarker
+	if currentBlock == 0 {
+		return 0
+	}
+	if currentBlock - blockNumMarker <= BLOCK_RANGE {
```
cyberliem: with Uint64, don't use - unless completely necessary to avoid panic. This can be rewritten as:
if currentBlock <= BLOCK_RANGE +blockNumMarker.

stat/fetcher.go
```diff
@@ -86,10 +111,97 @@ func (self *Fetcher) Run() error {
...
+		self.sleepTime = 10 * time.Minute
```
cyberliem: @halink0803 review this logic on varied sleeping time.

stat/storage/fee_setrate_bolt.go
```diff
@@ -0,0 +1,183 @@
...
+			if err != nil {
+				log.Printf("Cant convert %s to uint64", transaction.BlockNumber)
+				return err
+			}
+			txIndexUint, err := strconv.ParseUint(transaction.TransactionIndex, 10, 64)
+			if err != nil {
+				log.Printf("Cant convert %s to uint64", transaction.TransactionIndex)
+				return err
+			}
+			keyStoreUint := blockNumUint * 1000 + txIndexUint
```
cyberliem: make this scaling param (1000) a Const

stat/storage/fee_setrate_bolt.go
```diff
@@ -0,0 +1,183 @@
...
+	})
+	return seqFeeSetRate, err
+}
+
+func getFeeSetRate(c *bolt.Cursor, tickBlock, nextTickBlock, tickTime []byte) (common.FeeSetRate, error) {
+	sumFee := big.NewFloat(0)
+	gasInEther := big.NewFloat(0)
+	var feeSetRate common.FeeSetRate
+	
+	for k, v := c.Seek(tickBlock); k != nil && bytes.Compare(k, nextTickBlock) < 0; k, v = c.Next() {
```
cyberliem: Double check nextTickBlock ( it might be the record of the next day, since *bolt.Cusor.Seek() return next key if the key does not exist.

stat/fetcher.go
```diff
@@ -86,10 +113,97 @@ func (self *Fetcher) Run() error {
...
+		time.Sleep(self.sleepTime)
+	}
+}
+
+type APIResponse struct {
+	Message string                 `json:"message"`
+	Result  []common.SetRateTxInfo `json:"result"`
+}
+
+func (self *Fetcher) FetchTxs(client http.Client) {
```
cyberliem: This function should return an error, and handle err at the scope calling it.

PR #209, opened 2018-05-03
Export prune storage

cyberliem:
New design for archive package, now the whole core will use one S3archive object to backup Data.
The storage size controller is localized to each package: I.e, stat and reserve-data each will have their own storage controller to export, prune and backup data.


cyberliem:
[Push old data to S3 and prune them from the database (apply to all time series data)](https://trello.com/c/5d2fnBcX/216-push-old-data-to-s3-and-prune-them-from-the-database-apply-to-all-time-series-data)

PR #216, opened 2018-05-10
Fix trade logs skipping at full storage

cyberliem:
- Separate Check Duplicate logs from Store logs logic
- Handle when Db op store logs failed.


cyberliem:
[Fix: TradeLog syncing is not working properly when disk is full](https://trello.com/c/ki4KMhwJ/222-fix-tradelog-syncing-is-not-working-properly-when-disk-is-full)

PR #226, opened 2018-05-17
Price optimization and set rate messages

tranvictor:
Currently we use 50.1 gwei for all setrate operations. To improve it, we want to start with a low gas price and start increase the gas price with the same nonce if it is not mined intime.
The formula is:
if `n` == 0: `new_price` = base price
elsif `n` > 4: `new_price = old_price + 1`
else: `new_price = old_price + (50.1 - old_price)/(5-n)`
Where:
`new_price` is the new price we will use to replace old tx.
`old_price` is the old price we used with the old tx.
`n` is the number of txs we have tried for the nonce.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -146,6 +146,19 @@ func readablePrint(data map[ethereum.Address]byte) string {
 	return result
 }
 
+func allZero(a, b, c, d []*big.Int) bool {
+	all := append(a, b...)
+	all = append(all, c...)
+	all = append(all, d...)
+	big0 := big.NewInt(0)
+	for _, i := range all {
+		if i.Cmp(big0) != 0 {
+			return false
+		}
+	}
+	return true
+}
+
 //====================== Write calls ===============================
 
 // TODO: Need better test coverage
@@ -168,6 +181,15 @@ func (self *Blockchain) SetRates(
 	if err != nil {
 		return nil, err
 	}
+
+	// This is commented out because we dont want to make too much of change. Don't remove
+	// this check, it can be useful in the future.
+	//
+	// Don't submit any txs if it is just trying to set all tokens to 0 when they are already 0
+	// if allZero(buys, sells, baseBuys, baseSells) {
+	// 	return nil, errors.New("Trying to set all rates to 0 but they are already 0. Skip the tx.")
+	// }
+
 	baseTokens := []ethereum.Address{}
 	newBSells := []*big.Int{}
 	newBBuys := []*big.Int{}
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"math/big"
 	"strings"
+	"time"
 
 	"github.com/KyberNetwork/reserve-data/common"
 	"github.com/KyberNetwork/reserve-data/common/blockchain"
@@ -51,6 +52,9 @@ type Blockchain struct {
 	oldBurners    []ethereum.Address
 	tokens        []common.Token
 	tokenIndices  map[string]tbindex
+
+	localSetRateNonce     uint64
+	setRateNonceTimestamp uint64
 }
 
 func (self *Blockchain) AddOldNetwork(addr ethereum.Address) {
@@ -558,8 +562,33 @@ func (self *Blockchain) GetLogs(fromBlock uint64, toBlock uint64) ([]common.KNLo
 	return result, nil
 }
 
+// Getting mined nonce is not simple because there might be lag between
+// node leading us to get outdated mined nonce from an unsynced node.
+// To overcome this situation, we will keep a local nonce and require
+// the nonce from node to be equal or greater than it.
+// If the nonce from node is smaller than the local one, we will use
+// the local one. However, if the local one stay with the same value
+// for more than 15mins, the local one is considered incorrect
+// because the chain might be reorganized so we will invalidate it
+// and assign it to the nonce from node.
 func (self *Blockchain) SetRateMinedNonce() (uint64, error) {
-	return self.GetMinedNonce(PRICING_OP)
+	nonceFromNode, err := self.GetMinedNonce(PRICING_OP)
+	if err != nil {
+		return nonceFromNode, err
+	}
+	if nonceFromNode < self.localSetRateNonce {
+		if common.GetTimepoint()-self.setRateNonceTimestamp > uint64(15*time.Minute) {
+			self.localSetRateNonce = nonceFromNode
+			self.setRateNonceTimestamp = common.GetTimepoint()
+			return nonceFromNode, nil
+		} else {
+			return self.localSetRateNonce, nil
+		}
+	} else {
+		self.localSetRateNonce = nonceFromNode
+		self.setRateNonceTimestamp = common.GetTimepoint()
+		return nonceFromNode, nil
+	}
 }
 
 func NewBlockchain(
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -146,19 +146,6 @@ func readablePrint(data map[ethereum.Address]byte) string {
 	return result
 }
 
-func allZero(a, b, c, d []*big.Int) bool {
-	all := append(a, b...)
-	all = append(all, c...)
-	all = append(all, d...)
-	big0 := big.NewInt(0)
-	for _, i := range all {
-		if i.Cmp(big0) != 0 {
-			return false
-		}
-	}
-	return true
-}
-
 //====================== Write calls ===============================
 
 // TODO: Need better test coverage
@@ -186,7 +173,7 @@ func (self *Blockchain) SetRates(
 	// this check, it can be useful in the future.
 	//
 	// Don't submit any txs if it is just trying to set all tokens to 0 when they are already 0
-	// if allZero(buys, sells, baseBuys, baseSells) {
+	// if common.AllZero(buys, sells, baseBuys, baseSells) {
 	// 	return nil, errors.New("Trying to set all rates to 0 but they are already 0. Skip the tx.")
 	// }
 
@@ -584,6 +571,9 @@ func (self *Blockchain) GetLogs(fromBlock uint64, toBlock uint64) ([]common.KNLo
 	return result, nil
 }
 
+// SetRateMinedNonce returns nonce of the pricing operator in confirmed
+// state (not pending state).
+//
 // Getting mined nonce is not simple because there might be lag between
 // node leading us to get outdated mined nonce from an unsynced node.
 // To overcome this situation, we will keep a local nonce and require
--- a/common/blockchain/base_blockchain.go
+++ b/common/blockchain/base_blockchain.go
@@ -2,7 +2,6 @@ package blockchain
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log"
 	"math/big"
@@ -93,7 +92,7 @@ func (self *BaseBlockchain) GetNextNonce(operator string) (*big.Int, error) {
 func (self *BaseBlockchain) SignAndBroadcast(tx *types.Transaction, from string) (*types.Transaction, error) {
 	signer := self.GetOperator(from).Signer
 	if tx == nil {
-		panic(errors.New("Nil tx is forbidden here"))
+		panic(fmt.Errorf("Nil tx is forbidden here"))
 	} else {
 		signedTx, err := signer.Sign(tx)
 		if err != nil {
@@ -104,9 +103,9 @@ func (self *BaseBlockchain) SignAndBroadcast(tx *types.Transaction, from string)
 		if !ok {
 			log.Printf("Broadcasting transaction failed! nonce: %d, gas price: %s, retry failures: %s", tx.Nonce(), tx.GasPrice().Text(10), failures)
 			if signedTx != nil {
-				return signedTx, errors.New(fmt.Sprintf("Broadcasting transaction %s failed, retry failures: %s", tx.Hash().Hex(), failures))
+				return signedTx, fmt.Errorf("Broadcasting transaction %s failed, retry failures: %s", tx.Hash().Hex(), failures)
 			} else {
-				return signedTx, errors.New(fmt.Sprintf("Broadcasting transaction failed, retry failures: %s", failures))
+				return signedTx, fmt.Errorf("Broadcasting transaction failed, retry failures: %s", failures)
 			}
 		} else {
 			return signedTx, nil
@@ -167,13 +166,13 @@ func (self *BaseBlockchain) transactTx(context context.Context, opts TxOpts, con
 	}
 	var nonce uint64
 	if opts.Nonce == nil {
-		return nil, errors.New("nonce must be specified")
+		return nil, fmt.Errorf("nonce must be specified")
 	} else {
 		nonce = opts.Nonce.Uint64()
 	}
 	// Figure out the gas allowance and gas price values
 	if opts.GasPrice == nil {
-		return nil, errors.New("gas price must be specified")
+		return nil, fmt.Errorf("gas price must be specified")
 	}
 	gasLimit := opts.GasLimit
 	if gasLimit == nil {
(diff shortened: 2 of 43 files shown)


tranvictor:
[Gas price optimization for setrate opeartions](https://trello.com/c/Qc6MnRQe/11-gas-price-optimization-for-setrate-opeartions)

blockchain/blockchain.go
```diff
@@ -142,6 +146,19 @@ func readablePrint(data map[ethereum.Address]byte) string {
 	return result
 }
 
+func allZero(a, b, c, d []*big.Int) bool {
```
favadi: Should we make this function more generic by changing its signature to 

```golang
func allZero(...[]*big.Int) bool
```

Or we need a comment or reviewer will wonder why we accept 4 params, not 3 not 5.

blockchain/blockchain.go
```diff
@@ -164,6 +181,15 @@ func (self *Blockchain) SetRates(
 	if err != nil {
 		return nil, err
 	}
+
+	// This is commented out because we dont want to make too much of change. Don't remove
+	// this check, it can be useful in the future.
+	//
+	// Don't submit any txs if it is just trying to set all tokens to 0 when they are already 0
+	// if allZero(buys, sells, baseBuys, baseSells) {
```
favadi: uhm, so the allZero function is not used at all. Should it be commented out too?

blockchain/blockchain.go
```diff
@@ -558,8 +584,33 @@ func (self *Blockchain) GetLogs(fromBlock uint64, toBlock uint64) ([]common.KNLo
...
+		return nonceFromNode, err
+	}
+	if nonceFromNode < self.localSetRateNonce {
+		if common.GetTimepoint()-self.setRateNonceTimestamp > uint64(15*time.Minute) {
+			self.localSetRateNonce = nonceFromNode
+			self.setRateNonceTimestamp = common.GetTimepoint()
+			return nonceFromNode, nil
+		} else {
+			return self.localSetRateNonce, nil
+		}
```
favadi: It is common in Golang to use:

```golang
if err != nil {
	// error handling
	return // or continue, etc.
}
// normal code
```
instead of

```golang
if err != nil {
	// error handling
} else {
	// normal code
}
```

favadi: See: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

blockchain/blockchain.go
```diff
@@ -558,8 +584,33 @@ func (self *Blockchain) GetLogs(fromBlock uint64, toBlock uint64) ([]common.KNLo
...
+	} else {
+		self.localSetRateNonce = nonceFromNode
+		self.setRateNonceTimestamp = common.GetTimepoint()
+		return nonceFromNode, nil
```
favadi: same as above, see: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

common/blockchain/base_blockchain.go
```diff
@@ -102,11 +102,11 @@ func (self *BaseBlockchain) SignAndBroadcast(tx *types.Transaction, from string)
 		failures, ok := self.broadcaster.Broadcast(signedTx)
 		log.Printf("Rebroadcasting failures: %s", failures)
 		if !ok {
-			log.Printf("Broadcasting transaction failed!!!!!!!, err: %s, retry failures: %s", err, failures)
+			log.Printf("Broadcasting transaction failed! nonce: %d, gas price: %s, retry failures: %s", tx.Nonce(), tx.GasPrice().Text(10), failures)
```
favadi: maybe we really need to use structured logging soon.

common/blockchain/base_blockchain.go
```diff
@@ -278,7 +278,7 @@ func (self *BaseBlockchain) BuildSendERC20Tx(opts TxOpts, amount *big.Int, to et
 		return nil, err
 	}
 	msg := ether.CallMsg{From: opts.Operator.Address, To: &tokenAddress, Value: value, Data: data}
-	timeout, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	timeout, cancel := context.WithTimeout(context.Background(), 7*time.Second)
```
favadi: should this timeout is a const? I don't why we need to change it too.

common/blockchain/nonce/time_window.go
```diff
@@ -35,14 +35,14 @@ func (self *TimeWindow) GetAddress() ethereum.Address {
 }
 
 func (self *TimeWindow) getNonceFromNode(ethclient *ethclient.Client) (*big.Int, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 7*time.Second)
```
favadi: now I see the magic 7 seconds everywhere.

core/activity_storage.go
```diff
@@ -19,5 +19,6 @@ type ActivityStorage interface {
 
 	GetActivity(id common.ActivityID) (common.ActivityRecord, error)
 
-	PendingSetrate(minedNonce uint64) (*common.ActivityRecord, error)
+	// return the last pending set rate and number of pendint txs
+	PendingSetrate(minedNonce uint64) (*common.ActivityRecord, uint64, error)
```
favadi: The standard way to comment should be.

> PendingSetrate returns the last pending set rate and number of pending txs.

The reason is it will be generated as a complete sentence in https://godoc.org/github.com/KyberNetwork/reserve-data.

blockchain/blockchain.go
```diff
@@ -558,8 +584,33 @@ func (self *Blockchain) GetLogs(fromBlock uint64, toBlock uint64) ([]common.KNLo
 	return result, nil
 }
 
+// Getting mined nonce is not simple because there might be lag between
```
favadi: As we already care enough to add a comment, we should make the comment starts with `SetRateMinedNonce` word.

core/reserve_core.go
```diff
@@ -232,17 +232,34 @@ func (self ReserveCore) Withdraw(
...
-func (self ReserveCore) pendingSetrateInfo(minedNonce uint64) (*big.Int, *big.Int, error) {
-	act, err := self.activityStorage.PendingSetrate(minedNonce)
+func calculateNewGasPrice(old *big.Int, count uint64) *big.Int {
+	// in this case after 5 tries the tx is still not mined.
+	// at this point, 50.1 gwei is not enough but it doesn't matter
+	// if the tx is mined or not because users' tx is not mined neither
+	// so we can just increase the gas price a tiny amount (1 gwei) to make
+	// the node accept tx with up to date price
+	if count > 4 {
+		return old.Add(old, big.NewInt(1000000000))
```
favadi: I guess `big.NewInt(1000000000)` is equal to 1gwei. It is hard to count the zeros, should we move it to a package somewhere and reuse every times?

core/reserve_core.go
```diff
@@ -232,17 +232,34 @@ func (self ReserveCore) Withdraw(
...
+	} else {
+		// new = old + (50.1 - old) / (5 - count)
+		return old.Add(
+			old,
+			big.NewInt(0).Div(big.NewInt(0).Sub(big.NewInt(50100000000), old), big.NewInt(int64(5-count))),
```
favadi: Could you add a comment explain why we use the magic number `50100000000`? And should it be a const?

core/reserve_core.go
```diff
@@ -232,17 +232,34 @@ func (self ReserveCore) Withdraw(
...
+	// in this case after 5 tries the tx is still not mined.
+	// at this point, 50.1 gwei is not enough but it doesn't matter
+	// if the tx is mined or not because users' tx is not mined neither
+	// so we can just increase the gas price a tiny amount (1 gwei) to make
+	// the node accept tx with up to date price
+	if count > 4 {
+		return old.Add(old, big.NewInt(1000000000))
+	} else {
+		// new = old + (50.1 - old) / (5 - count)
+		return old.Add(
```
favadi: same comment a bout if error flow like above.

core/reserve_core.go
```diff
@@ -232,17 +232,34 @@ func (self ReserveCore) Withdraw(
...
+		return nil, nil, 0, err
 	}
 	if act != nil {
 		nonce, _ := strconv.ParseUint(act.Result["nonce"].(string), 10, 64)
 		gasPrice, _ := strconv.ParseUint(act.Result["gasPrice"].(string), 10, 64)
-		return big.NewInt(int64(nonce)), big.NewInt(int64(gasPrice)), nil
+		return big.NewInt(int64(nonce)), big.NewInt(int64(gasPrice)), count, nil
 	} else {
-		return nil, nil, nil
+		return nil, nil, 0, nil
```
favadi: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

core/reserve_core.go
```diff
@@ -278,27 +296,30 @@ func (self ReserveCore) SetRates(
...
-						newPrice := big.NewInt(0).Add(oldPrice, big.NewInt(10000000000))
+						newPrice := calculateNewGasPrice(oldPrice, count)
 						log.Printf("Trying to replace old tx with new price: %s", newPrice.Text(10))
 						tx, err = self.blockchain.SetRates(
 							tokenAddrs, buys, sells, block,
 							oldNonce,
 							newPrice,
 						)
 					} else {
+						initPrice := big.NewInt(10000000000)
```
favadi: it should be a const and with a comment explain int human readable unit (wei, gwei, ether...)

data/fetcher/fetcher.go
```diff
@@ -136,13 +136,18 @@ func (self *Fetcher) FetchRate(timepoint uint64) {
...
-	err = self.storage.StoreRate(data, timepoint)
-	// fmt.Printf("balance data: %v\n", data)
-	if err != nil {
-		log.Printf("Storing rates failed: %s", err.Error())
+		log.Printf("Fetching rates from blockchain failed: %s. Will not store it to storage.", err.Error())
+	} else {
+		if data.Valid {
+			log.Printf("Got rates from blockchain: %+v", data)
+			err = self.storage.StoreRate(data, timepoint)
+			// fmt.Printf("balance data: %v\n", data)
```
favadi: please remove this unused line.

interfaces.go
```diff
@@ -51,6 +51,9 @@ type ReserveData interface {
 	CurrentAuthDataVersion(timestamp uint64) (common.Version, error)
 	GetAuthData(timestamp uint64) (common.AuthDataResponse, error)
 
+	// all rates here must be valid rates. A rate is considered valid
```
favadi: Comment should be starts with `CurrentRateVersion`.

blockchain/blockchain.go
cyberliem: The logic flow should be : 
 
```go
if err !=nil {
	return
}

if noneFromNode < self.localSetRateNonce && common.GetTimepoint()-self.setRateNonceTimeStamp> uint64 (15*time.*minute) {
	return self.localSetRateNonce, nil
} 

self.localSetRateNonce = nonceFromNode
self.setRateNonceTimestamp = common.GetTimepoint()
return nonceFromNode, nil

```

blockchain/blockchain.go
tranvictor: Yeah agree

blockchain/blockchain.go
tranvictor: It should be, i will comment it.

common/blockchain/base_blockchain.go
tranvictor: Agree, but lets leave it for later improvement. Right now it is too much.

common/blockchain/nonce/time_window.go
tranvictor: 7s is really a magic number. I'm not sure we should make it a constant, it is just something that we tune along production (very rarely) and it is easy for reader, they read and know it is 7s timeout.

core/activity_storage.go
tranvictor: Yeah

core/reserve_core.go
tranvictor: We should

core/reserve_core.go
tranvictor: Ok will add some

blockchain/blockchain.go
tranvictor: First this is not error handling.
Second, the way I express the logic by code is intuitive and more readable even though it is longer than Liem's one so I will leave it as is.

blockchain/blockchain.go
tranvictor: I will leave it as is here too.

blockchain/blockchain.go
favadi: I thought it is error handling. The reason we prefer the style 

```go
if err != nil {
	// error handling
	return // or continue, etc.
}
// normal code
```

is because it encourages proper error handling by handle the worst case first and leave the normal path in the not-error-handling codes indent level.

So I have no objection if it is not error handling.

common/blockchain/base_blockchain.go
```diff
@@ -102,11 +102,11 @@ func (self *BaseBlockchain) SignAndBroadcast(tx *types.Transaction, from string)
 		failures, ok := self.broadcaster.Broadcast(signedTx)
 		log.Printf("Rebroadcasting failures: %s", failures)
 		if !ok {
-			log.Printf("Broadcasting transaction failed!!!!!!!, err: %s, retry failures: %s", err, failures)
+			log.Printf("Broadcasting transaction failed! nonce: %d, gas price: %s, retry failures: %s", tx.Nonce(), tx.GasPrice().Text(10), failures)
 			if signedTx != nil {
-				return signedTx, errors.New(fmt.Sprintf("Broadcasting transaction %s failed, err: %s, retry failures: %s", tx.Hash().Hex(), err, failures))
+				return signedTx, errors.New(fmt.Sprintf("Broadcasting transaction %s failed, retry failures: %s", tx.Hash().Hex(), failures))
```
favadi: we should use `fmt.Errorf` instead of `errors.New`.

common/blockchain/base_blockchain.go
```diff
@@ -102,11 +102,11 @@ func (self *BaseBlockchain) SignAndBroadcast(tx *types.Transaction, from string)
...
 			} else {
-				return signedTx, errors.New(fmt.Sprintf("Broadcasting transaction failed, err: %s, retry failures: %s", err, failures))
+				return signedTx, errors.New(fmt.Sprintf("Broadcasting transaction failed, retry failures: %s", failures))
```
favadi: we should use `fmt.Errorf` instead of `errors.New`.

core/reserve_core.go
```diff
@@ -232,26 +236,43 @@ func (self ReserveCore) Withdraw(
...
-		nonce, _ := strconv.ParseUint(act.Result["nonce"].(string), 10, 64)
-		gasPrice, _ := strconv.ParseUint(act.Result["gasPrice"].(string), 10, 64)
-		return big.NewInt(int64(nonce)), big.NewInt(int64(gasPrice)), nil
-	} else {
-		return nil, nil, nil
+	if act == nil {
+		return nil, nil, 0, nil
 	}
+	nonce, _ := strconv.ParseUint(act.Result["nonce"].(string), 10, 64)
+	gasPrice, _ := strconv.ParseUint(act.Result["gasPrice"].(string), 10, 64)
```
favadi: should we handle the errors here?

data/fetcher/fetcher.go
```diff
@@ -136,13 +136,17 @@ func (self *Fetcher) FetchRate(timepoint uint64) {
...
-	err = self.storage.StoreRate(data, timepoint)
-	// fmt.Printf("balance data: %v\n", data)
-	if err != nil {
-		log.Printf("Storing rates failed: %s", err.Error())
+		log.Printf("Fetching rates from blockchain failed: %s. Will not store it to storage.", err.Error())
+	} else {
+		if data.Valid {
+			log.Printf("Got rates from blockchain: %+v", data)
+			err = self.storage.StoreRate(data, timepoint)
+			if err != nil {
```
favadi: minor, but we should use `if err = xxx; err != nil {}` style to make sure we never forgot to handle the returned error.

core/reserve_core.go
tranvictor: We can ignore the error here because the nonce and gas price is enforced in other scope already.

data/fetcher/fetcher.go
tranvictor: good point

common/blockchain/base_blockchain.go
```diff
@@ -93,7 +92,7 @@ func (self *BaseBlockchain) GetNextNonce(operator string) (*big.Int, error) {
 func (self *BaseBlockchain) SignAndBroadcast(tx *types.Transaction, from string) (*types.Transaction, error) {
 	signer := self.GetOperator(from).Signer
 	if tx == nil {
-		panic(errors.New("Nil tx is forbidden here"))
+		panic(fmt.Errorf("Nil tx is forbidden here"))
```
favadi: actually, we should use `errors.New` when there is no formatting directive and `fmt.Errorf` when there is. Reason is errors.New is multiple times faster than fmt.Errorf. Another reason is golint will complain about this and reduce our score in Go Report Card :nerd_face: .

tranvictor: ah ok, good to know about our scoring concern

PR #227, opened 2018-05-17
Update set target qty v2

halink0803:


halink0803:
[Update set token target quantity api v2 for analytics](https://trello.com/c/koKLpyAQ/18-update-set-token-target-quantity-api-v2-for-analytics)

data/storage/bolt.go
```diff
@@ -1428,3 +1438,131 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+		return nil
+	})
+	if len(result) == 0 {
+		targetQty, err := self.GetTokenTargetQty()
+		if err != nil {
+			return result, err
+		}
+		strTargets := strings.Split(targetQty.Data, "|")
+		log.Printf("string target: %+v", strTargets)
+		for index, target := range strTargets {
```
halink0803: @favadi please help me check this loop, when strTarget is empty, this loop still come inside?
Steps to reproduce panic error:
- remove cmd/dev.db
- run core --noauth
- go to http://localhost:8000/targetqty/v2

halink0803:
@cyberliem That CI error that @favadi mentioned he is missing in the task https://trello.com/c/UHc5gled/10-commonexchangegetinfo-is-returning-a-copy-of-mutex-lock

I don't know if I should resolve it here or let it in other task.

README.md
```diff
@@ -1212,6 +1212,115 @@ response:
   {"data":{"Timestamp":1524852506656,"DGX":{"Valid":false,"Timestamp":0,"success":"","data":[{"symbol":"ETHDGX","price":0.07238485,"time":1524852506},{"symbol":"ETHUSD","price":713,"time":1524852506},{"symbol":"ETHSGD","price":944,"time":1524852506},{"symbol":"DGXUSD","price":49,"time":1524852506},{"symbol":"EURUSD","price":1.21063,"time":1524852505},{"symbol":"USDSGD","price":1.32439,"time":1524852505},{"symbol":"XAUUSD","price":1322.62,"time":1524852505},{"symbol":"USDJPY","price":109.105,"time":1524852505}],"Error":""},"OneForge":{"Value":1.9465,"Text":"1 XAU is worth 1.9465 ETH","Timestamp":1524852506,"Error":false,"Message":""}},"success":true}
 ```
 
+
+### set target quantity v2- (signing required)
+```
+<host>:8000/settargetqty/v2
```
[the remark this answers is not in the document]
halink0803: @favadi not having a proper reason, I just named the api that as it might easy to remember.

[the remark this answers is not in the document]
favadi: I think the `/endpoint/v2` scheme is more common: http://restcookbook.com/Basics/versioning/.

@tranvictor your thought?

data/storage/bolt.go
```diff
@@ -45,6 +45,8 @@ const (
 	STABLE_TOKEN_PARAMS_BUCKET         string = "stable-token-params"
 	PENDING_STABLE_TOKEN_PARAMS_BUCKET string = "pending-stable-token-params"
 	GOLD_BUCKET                        string = "gold_feeds"
+	PENDING_TARGET_QUANTITY_V2         string = "pending_target_qty_v2"
```
favadi: @halink0803 could you add one comment for each new const? It help reviewer to understand the PR better.

data/storage/bolt.go
```diff
@@ -143,6 +145,14 @@ func NewBoltStorage(path string) (*BoltStorage, error) {
 		if err != nil {
 			return err
 		}
+		_, err = tx.CreateBucketIfNotExists([]byte(PENDING_TARGET_QUANTITY_V2))
+		if err != nil {
```
favadi:
```golang
_, err = xxx
if err != nil {
}
```

should be replaced with

```golang
if _, err = xxx; err != nil {
}
```

data/storage/bolt.go
```diff
@@ -143,6 +145,14 @@ func NewBoltStorage(path string) (*BoltStorage, error) {
...
+			return err
+		}
+		_, err = tx.CreateBucketIfNotExists([]byte(TARGET_QUANTITY_V2))
+		if err != nil {
```
favadi: same as above.

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+
+func (self *BoltStorage) StorePendingTargetQtyV2(value []byte) error {
+	k := uint64ToBytes(1)
+	temp := make(map[string]interface{})
+	vErr := json.Unmarshal(value, &temp)
+	if vErr != nil {
+		return fmt.Errorf("Rejected: Data could not be unmarshalled to defined format: %s", vErr.Error())
+	}
+	err := self.db.Update(func(tx *bolt.Tx) error {
+		b, uErr := tx.CreateBucketIfNotExists([]byte(PENDING_TARGET_QUANTITY_V2))
```
favadi: do we really need to check of bucket exists every time we use it?

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+	k := uint64ToBytes(1)
+	result := make(map[string]interface{})
+	err := self.db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(PENDING_TARGET_QUANTITY_V2))
+		if b == nil {
+			return fmt.Errorf("Bucket hasn't exist yet")
+		}
+		record := b.Get(k)
+		if record == nil {
+			return nil
```
favadi: Should we returns an error not found here or it is the same if:

- no pending target is stored
- there is a pending target is stored with empty value `{}`

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+	vErr := json.Unmarshal(value, &temp)
+	if vErr != nil {
+		return fmt.Errorf("Rejected: Data could not be unmarshalled to defined format: %s", vErr)
+	}
+	pending, err := self.GetPendingTargetQtyV2()
+	if eq := reflect.DeepEqual(pending, temp); !eq {
+		return fmt.Errorf("Rejected: confiming data isn't consistent")
+	}
+
+	err = self.db.Update(func(tx *bolt.Tx) error {
```
favadi: So the `GetPendingTargetQtyV2` above and this are running in two separate transactions. To be safe they are should be running in one transaction, or following scenarios might happen:
- user called `ConfirmTargetQtyV2`
- `ConfirmTargetQtyV2` get pending quantity in L1492
- before L1497 is executed, user called `RemovePendingTargetQtyV2`. The operation is completed before `ConfirmTargetQtyV2` reaches L1497.
- L1497 is executed, but now the value is removed from pending bucket
- L1507 will return an error

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+		b, uErr := tx.CreateBucketIfNotExists([]byte(TARGET_QUANTITY_V2))
+		if uErr != nil {
+			return uErr
+		}
+		return b.Put(k, value)
+	})
+	if err != nil {
+		return err
+	}
+	err = self.RemovePendingTargetQtyV2()
```
favadi: same concern about transaction as above. I think all three database operation should be in one transaction with proper rollback/commit

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+	k := uint64ToBytes(1)
+	result := make(map[string]interface{})
+	err := self.db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(TARGET_QUANTITY_V2))
+		if b == nil {
+			return fmt.Errorf("Bucket hasn't exist yet")
+		}
+		record := b.Get(k)
+		if record == nil {
+			return nil
```
favadi: same question about if we need to return not found error here.

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+func (self *BoltStorage) RemovePendingTargetQtyV2() error {
+	k := uint64ToBytes(1)
+	err := self.db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(PENDING_TARGET_QUANTITY_V2))
+		if b == nil {
+			return fmt.Errorf("Bucket hasn't existed yet")
+		}
+		record := b.Get(k)
+		if record == nil {
+			return fmt.Errorf("Bucket is empty")
```
favadi: uhm, is it really an error? Normally, /DELETE a non existing resource is not considered an error most of the time.

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+		if record == nil {
+			return nil
+		}
+		vErr := json.Unmarshal(record, &result)
+		if vErr != nil {
+			return vErr
+		}
+		return nil
+	})
+	if len(result) == 0 {
```
favadi: could you add a comment like, what do that blocks do when the result is empty?

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+		strTargets := strings.Split(targetQty.Data, "|")
+		for _, target := range strTargets {
+			elements := strings.Split(target, "_")
+			if len(elements) != 5 {
+				continue
+			}
+			totalTarget, _ := strconv.ParseFloat(elements[1], 10)
+			reserveTarget, _ := strconv.ParseFloat(elements[2], 10)
+			rebalance, _ := strconv.ParseFloat(elements[3], 10)
+			withdraw, _ := strconv.ParseFloat(elements[4], 10)
```
favadi: I can't verify if those string operations are correct or not. Could you refactor them to a function and comment with example input, output? Better if you add some test cases to that function as well.

http/server.go
```diff
@@ -2577,6 +2575,152 @@ func (self *HTTPServer) GetTokenHeatmap(c *gin.Context) {
...
 
+func (self *HTTPServer) SetTargetQtyV2(c *gin.Context) {
+	postForm, ok := self.Authenticated(c, []string{}, []Permission{ConfigurePermission})
+	if !ok {
+		return
+	}
+	value := []byte(postForm.Get("value"))
+	if len(value) > MAX_DATA_SIZE {
+		c.JSON(
+			http.StatusOK,
```
favadi: @tranvictor @halink0803 just to confirm, so we don't use standard http status code but return 200 even if the operation failed

halink0803: @favadi when the operation failed, we will return the reason back to client.

data/storage/bolt.go
halink0803: @favadi The pending target should not be empty {}, so it returns {}, it means there is no pending target.

data/storage/bolt.go
halink0803: @favadi I agree, let me fix this.

http/server.go
favadi: > so we don't use standard http status code but return 200 even if the operation failed

It is a "yes", "no" question.

data/storage/bolt.go
favadi: If we never push anything to pending bucket, it should be empty, right?

data/storage/bolt.go
```diff
@@ -1428,3 +1438,132 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+			return fmt.Errorf("Currently there is a pending record")
+		}
+		log.Printf("Token target qty: %s", value)
+		return b.Put(k, value)
+	})
+	return err
+}
+
+func (self *BoltStorage) GetPendingTargetQtyV2() (map[string]interface{}, error) {
+	k := uint64ToBytes(1)
```
favadi: I miss this on first review. Should we use a meaningful key name instead?

data/storage/bolt.go
halink0803: I agree.

data/storage/bolt.go
halink0803: @favadi what do you think is a meaningful key name here? This is a one time record and we always overwrite it when having new value.

http/server.go
halink0803: Yes, we return 200 even if the operation failed.

README.md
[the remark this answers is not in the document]
halink0803: @favadi @tranvictor So, finally should I change it to /v2/endpoint or keep it /endpoint/v2?

data/storage/bolt.go
favadi: anything but "1", something like "current_pending_target_quantity" should do it.

README.md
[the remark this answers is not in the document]
tranvictor: I'm neutral. Lets just decide on something and keep it consistently though

data/storage/bolt.go
tranvictor: I support @favadi, it is nicer

data/storage/bolt.go
tranvictor: This is valid but lets make another issue for it. The issue will not happen in practice because there are only 1/2 person using it on deployment

data/storage/bolt.go
tranvictor: as long as it is consistent, we can go with error. I do think it should return empty, it makes more sense. Please make another issue for this to do later.

http/server.go
tranvictor: For now we are ok with it. Chaning it will affect other components too, lets have something about this in the backlogs

data/storage/bolt.go
favadi: Task created: https://trello.com/c/o2j5MzwR/34-boltdb-ensure-that-multiple-operations-is-protected-in-a-single-transaction

data/storage/bolt.go
favadi: https://trello.com/c/78Ix4gRT/35-boltdb-review-get-operation-handling-for-not-exist-object

data/storage/bolt.go
```diff
@@ -1428,3 +1443,135 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+
+func (self *BoltStorage) GetPendingTargetQtyV2() (map[string]interface{}, error) {
+	result := make(map[string]interface{})
+	err := self.db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(PENDING_TARGET_QUANTITY_V2))
+		k := []byte("current_pending_target_qty")
+		record := b.Get(k)
+		if record != nil {
+			vErr := json.Unmarshal(record, &result)
+			if vErr != nil {
```
favadi: please use use `if err := xxx; err != nil {}` form.

data/storage/bolt.go
```diff
@@ -1428,3 +1443,135 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+	return err
+}
+
+// RemovePendingTargetQtyV2 remove pending data from db
+func (self *BoltStorage) RemovePendingTargetQtyV2() error {
+	err := self.db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(PENDING_TARGET_QUANTITY_V2))
+		if b == nil {
+			return fmt.Errorf("Bucket hasn't existed yet")
+		}
```
favadi: @halink0803 I think I still have an unanswered question: wheter if we need to check bucket exists in every operation or not. What do you think?

data/storage/bolt.go
```diff
@@ -1428,3 +1443,135 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+		k := []byte("current_target_qty")
+		record := b.Get(k)
+		if record == nil {
+			return nil
+		}
+		return json.Unmarshal(record, &result)
+	})
+
+	// This block below ís for backward compatible for api v1
```
favadi: this "is" word has a typo, right?

data/storage/bolt.go
```diff
@@ -1428,3 +1443,135 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+	// when the result is empty it means there is not target quantity is set
+	// we need to get current target quantity from v1 bucket and return it as v2 form.
+	if len(result) == 0 {
+
```
favadi: please remove this empty line.

data/storage/bolt.go
```diff
@@ -1428,3 +1443,135 @@ func (self *BoltStorage) RemovePendingStableTokenParams() error {
...
+	return result, err
+}
+
+// This function convert target quantity from v1 to v2
+// TokenTargetQty v1 should be follow this format:
+// token_totalTarget_reserveTarget_rebalanceThreshold_transferThreshold|token_totalTarget_reserveTarget_rebalanceThreshold_transferThreshold|...
+// while token is a string, it is validated before it saved then no need to validate again here
+// totalTarget, reserveTarget, rebalanceThreshold and transferThreshold are float numbers
+// and they are also no need to check to error here also (so we can ignore as below)
+func convertTargetQtyV1toV2(target metric.TokenTargetQty) map[string]interface{} {
```
favadi: It would be best if there is a a test for this function, as it is quite hard for me to know if it is working correctly looking at the codes.

http/server.go
```diff
@@ -2577,6 +2575,152 @@ func (self *HTTPServer) GetTokenHeatmap(c *gin.Context) {
...
+		)
+		return
+	}
+	c.JSON(
+		http.StatusOK,
+		gin.H{
+			"success": true,
+			"data":    data,
+		},
+	)
```
favadi: you can reduce code duplication by doing something like that:

```golang


if err != nil {
 resp = gin.H{
    "success": false,
    "reason": err.Error(),
  }   
} else {
resp := gin.H {
    "success": true,
    "data": data
}
}

c.JSON(http.StatusOK, resp)

```

Just cosmetic details though.

favadi: ANW, I created this https://trello.com/c/ss79sz7B/38-codes-duplication-in-http-handlers

http/server.go
```diff
@@ -2611,6 +2755,12 @@ func (self *HTTPServer) Run() {
 		self.r.POST("/confirmtargetqty", self.ConfirmTargetQty)
 		self.r.POST("/canceltargetqty", self.CancelTargetQty)
 
+		self.r.GET("/v2/targetqty", self.GetTargetQtyV2)
+		self.r.GET("/v2/pendingtargetqty", self.GetPendingTargetQtyV2)
+		self.r.POST("/v2/settargetqty", self.SetTargetQtyV2)
+		self.r.POST("/v2/confirmtargetqty", self.ConfirmTargetQtyV2)
+		self.r.POST("/v2/canceltargetqty", self.CancelTargetQtyV2)
```
favadi: Please use router group feature.

```
v2 := router.Group("/v2")
v2.Get(xxx,yyy)
```

metric/types.go
```diff
@@ -46,3 +46,10 @@ type RebalanceControl struct {
 type SetrateControl struct {
 	Status bool `json:"status"`
 }
+
+type TargetQtyStruct struct {
+	TotalTarget        float64 `json:"TotalTarget"`
+	ReserveTarget      float64 `json:"ReseveTarget"`
+	RebalanceThreshold float64 `json:"RebalanceThreshold"`
+	TransferThreshold  float64 `json:"TransferThreshold"`
```
favadi: those json tags look weird, most of the time I saw them in form "total_target", not "TotalTarget". Could you check which form we use in most of our code base.

data/storage/bolt.go
halink0803: @favadi there are two kind of bucket, one is that which initiate from NewXXXStorage function, this kind of bucket does not need to check bucket exists (this bucket is one of that kind). The other kind of bucket is nested bucket, which is created under parents bucket while the tx happen, this kind of bucket need to check if exists. You can find this kind of bucket in stat storage.

data/storage/bolt.go
halink0803: Yes, will fix it.

metric/types.go
halink0803: Yep, I agree. Those json tags look weird, however those json tags are follow a form from analytic, I will discuss with them to standardize this.

data/storage/bolt.go
favadi: Thanks for explanation.

> this kind of bucket does not need to check bucket exists (this bucket is one of that kind)

I still see it is checked for existent above, does it need to be fixed?

data/storage/bolt.go
halink0803: Removed it.

data/storage/bolt.go
halink0803: Yep, applied.

http/server.go
halink0803: Applied.

metric/types.go
halink0803: As discuss, I will keep it as it is because we make it consistent between components.

data/storage/bolt.go
halink0803: I write test for this.

data/storage/bolt.go
halink0803: I removed the check for existent.

data/storage/bolt.go
favadi: uhm, it is still here, do you forget to push?

PR #235, opened 2018-05-18
fix GetInfo is returning copy of mutex lock

favadi:


favadi:
[common.Exchange.GetInfo() is returning a copy of mutex lock](https://trello.com/c/UHc5gled/10-commonexchangegetinfo-is-returning-a-copy-of-mutex-lock)

PR #240, opened 2018-05-21
update from id for binance trade history

halink0803:
I have misreading the binance API, if not provide fromId params, binance api will return most recent records, instead of return from 0.

I changed it, set default to 0. It should run correctly now.

### DIFF
--- a/exchange/binance/binance_endpoint.go
+++ b/exchange/binance/binance_endpoint.go
@@ -195,6 +195,8 @@ func (self *BinanceEndpoint) GetAccountTradeHistory(
 	}
 	if fromID != "" {
 		params["fromId"] = fromID
+	} else {
+		params["fromId"] = "0"
 	}
 	resp_body, err := self.GetResponse(
 		"GET",


exchange/binance/binance_endpoint.go
```diff
@@ -195,6 +195,8 @@ func (self *BinanceEndpoint) GetAccountTradeHistory(
 	}
 	if fromID != "" {
 		params["fromId"] = fromID
+	} else {
+		params["fromId"] = "0"
```
favadi: according to the Binance API docs: https://github.com/binance-exchange/binance-official-api-docs/blob/master/rest-api.md#account-trade-list-user_data.

The `fromID` is not mandatory. Could you explain why set it to 0 make a different? And what 0 means in this case?

favadi:
[Fix get exchange trade history](https://trello.com/c/fthmdKUz/24-fix-get-exchange-trade-history)

exchange/binance/binance_endpoint.go
halink0803: @favadi We want to retrieve all the trade history from centralized exchange (in this case it is binance). From id is the last id which we retrieved history, if it is empty, it means that the first time we get history then we should set it to 0.

PR #244, opened 2018-05-21
get gas from gas station

NNagato:


tranvictor:
[Using gas station to estimate gas price to init on setrate tx](https://trello.com/c/mDWxYL26/37-using-gas-station-to-estimate-gas-price-to-init-on-setrate-tx)

PR #253, opened 2018-05-23
Add httputil package

favadi:


http/server.go
```diff
@@ -1208,54 +854,28 @@ func (self *HTTPServer) GetTradeHistory(c *gin.Context) {
...
 		history, err := ex.GetTradeHistory(fromTime, toTime)
 		if err != nil {
-			c.JSON(
-				http.StatusOK,
-				gin.H{
-					"success": false,
-					"data":    err.Error(),
-				},
-			)
+			// TODO: response should include error in reason field instead of data?
```
favadi: @halink0803 could you help me with this TODO? As I read the rest, it should return the error value in reason key instead of data.

favadi:
[Codes duplication in HTTP handlers](https://trello.com/c/ss79sz7B/38-codes-duplication-in-http-handlers)

http/server.go
halink0803: I changed it. It should be return reason instead of data.

PR #255, opened 2018-05-23
add nonce validator

favadi:


favadi:
[Mark tx as failed in some special cases](https://trello.com/c/RjkoE48R/43-mark-tx-as-failed-in-some-special-cases)

PR #257, opened 2018-05-24
create testing database file in tmp dir

favadi:


favadi:
[Temp files are created in source directory](https://trello.com/c/8chVKUaw/45-temp-files-are-created-in-source-directory)

cmd/configuration/user_storage_test.go
```diff
@@ -33,7 +33,11 @@ func doOneTest(f func(tester *stat.UserStorageTest, t *testing.T), t *testing.T)
 	if err != nil {
 		t.Fatalf("Testing bolt as a stat storage: init failed(%s)", err)
 	}
-	defer tearDownFn()
+	defer func() {
```
cyberliem: I see this func everywhere. Maybe we should create a testUtil subpackage to handle this?

favadi: These few lines look quite simple to me, not sure it is worth creating a new package just for them. I'm fine with either, so it is up to @halink0803 .

halink0803: It could be in testutil package, however as @favadi said, it quite too small so let move it in testutil in another PR when we have more feature to put it it. Are you okay with that? @cyberliem

PR #258, opened 2018-05-24
Unused codes

favadi:


favadi:
[Cleanup unused codes](https://trello.com/c/C1OfOYss/51-cleanup-unused-codes)

common/bolt.go
```diff
@@ -0,0 +1,21 @@
+package common
```
tranvictor: I dont think putting bolt specific utils in common is appropriate. In my initial ideal, common is a package that defines the types and utils to be shared among highest level components (core, data, stat).
To those packages, they always see abstracted storage with specific interfaces, not bolt.
**Proposal:** I think an util file inside of bolt specific package makes more sense.

favadi: I did that in 45481ab.

cmd/clicmd/startserver.go
```diff
@@ -24,49 +20,12 @@ const (
...
 var servPort int = 8000
 var endpointOW string
-var base_url, auth_url string
+var base_url string
 var enableStat bool
 var noCore bool
 var stdoutLog bool
 var dryrun bool
 
-func backupLog(arch archive.Archive) {
```
cyberliem: This function is necessary. Don't delete it. The function call has been accidentally  deleted

favadi: @cyberliem so.. the logging backup feature has been broken for a while? Could you push a commit to revert my changes to this function and add the call back?

cyberliem: I update my latest commit

tranvictor: @favadi it is not deployed to production nor staging yet so don't worry.

PR #259, opened 2018-05-25
Boltdb transaction

favadi:


favadi:
[BoltDB: ensure that multiple operations is protected in a single transaction](https://trello.com/c/o2j5MzwR/34-boltdb-ensure-that-multiple-operations-is-protected-in-a-single-transaction)

PR #261, opened 2018-05-28
change exchange trade history

halink0803:
Move include exchanges into reserve data instead of put it in http object.

### DIFF
--- a/data/reserve_data.go
+++ b/data/reserve_data.go
@@ -308,7 +308,6 @@ func (self ReserveData) ControlAuthDataSize() error {
 }
 
 func (self ReserveData) GetTradeHistory(fromTime, toTime uint64) (common.AllTradeHistory, error) {
-	var err error
 	data := common.AllTradeHistory{
 		Timestamp: common.GetTimestamp(),
 		Data:      map[common.ExchangeID]common.ExchangeTradeHistory{},
@@ -320,7 +319,7 @@ func (self ReserveData) GetTradeHistory(fromTime, toTime uint64) (common.AllTrad
 		}
 		data.Data[ex.ID()] = history
 	}
-	return data, err
+	return data, nil
 }
 
 func (self ReserveData) RunStorageController() error {
--- a/data/reserve_data.go
+++ b/data/reserve_data.go
@@ -309,16 +309,15 @@ func (self ReserveData) ControlAuthDataSize() error {
 
 func (self ReserveData) GetTradeHistory(fromTime, toTime uint64) (common.AllTradeHistory, error) {
 	data := common.AllTradeHistory{}
+	data.Data = map[common.ExchangeID]common.ExchangeTradeHistory{}
 	for _, ex := range self.exchanges {
 		history, err := ex.GetTradeHistory(fromTime, toTime)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


halink0803:
[Change api history design](https://trello.com/c/BCqSgwjM/40-change-api-history-design)

data/reserve_data.go
```diff
@@ -306,16 +307,35 @@ func (self ReserveData) ControlAuthDataSize() error {
...
+func (self ReserveData) GetTradeHistory(fromTime, toTime uint64) (common.AllTradeHistory, error) {
+	var err error
+	data := common.AllTradeHistory{
+		Timestamp: common.GetTimestamp(),
+		Data:      map[common.ExchangeID]common.ExchangeTradeHistory{},
+	}
+	for _, ex := range self.exchanges {
+		history, err := ex.GetTradeHistory(fromTime, toTime)
+		if err != nil {
+			return data, err
```
favadi: should we return nil, err? Because we should never evaluate data if err != nil.

data/reserve_data.go
```diff
@@ -306,16 +307,35 @@ func (self ReserveData) ControlAuthDataSize() error {
...
 }
 
+func (self ReserveData) GetTradeHistory(fromTime, toTime uint64) (common.AllTradeHistory, error) {
+	var err error
+	data := common.AllTradeHistory{
+		Timestamp: common.GetTimestamp(),
+		Data:      map[common.ExchangeID]common.ExchangeTradeHistory{},
+	}
+	for _, ex := range self.exchanges {
+		history, err := ex.GetTradeHistory(fromTime, toTime)
```
favadi: this err instance will shadow the err instance in L311. Please remove L311 as you don't assign it to anything in this function.

http/server.go
```diff
@@ -846,17 +845,9 @@ func (self *HTTPServer) GetTradeHistory(c *gin.Context) {
...
-	for _, ex := range self.exchanges {
-		history, err := ex.GetTradeHistory(fromTime, toTime)
-		if err != nil {
-			httputil.ResponseFailure(c, httputil.WithError(err))
-			return
-		}
-		data.Data[ex.ID()] = history
+	data, err := self.app.GetTradeHistory(fromTime, toTime)
+	if err != nil {
+		httputil.ResponseFailure(c, httputil.WithReason(err.Error()))
```
favadi: you added the httputil.WithError option right? Please use it. :)

data/reserve_data.go
halink0803: We have define return AllTradeHistory. Cannot return nil, plus we also check err before evaluate data then it should have encountered error here.

data/reserve_data.go
halink0803: Yep, done.

http/server.go
halink0803: Yep, I used it. Actually I did not add WithError function.

data/reserve_data.go
favadi: @halink0803 In L311, I see that data is initialized with non zero value.

halink0803: Saw it. Change it to zero value.

PR #268, opened 2018-05-30
Reorganize core setting token db path

favadi:


favadi:
[Persist Supported Token, separate it to a general package in setting](https://trello.com/c/YmslbFCg/20-persist-supported-token-separate-it-to-a-general-package-in-setting)

PR #272, opened 2018-05-31
update check deep equal for data confirmation

halink0803:


halink0803:
[Update check deep equal confirmation data](https://trello.com/c/3CcNJB4z/75-update-check-deep-equal-confirmation-data)

PR #274, opened 2018-05-31
update dev settings and json tag

halink0803:


halink0803:
[Fix json tag](https://trello.com/c/zYcCShWk/76-fix-json-tag)

PR #283, opened 2018-06-04
use enums for address/address set names

favadi:


favadi:
[Add address setting to setting package.](https://trello.com/c/fL9ivLQu/59-add-address-setting-to-setting-package)

PR #287, opened 2018-06-05
fix failed to store rate on a blank database

favadi:


favadi:
[Develop Bug: Storing rates failed.](https://trello.com/c/C1YQgcY8/93-develop-bug-storing-rates-failed)

data/storage/bolt.go
```diff
@@ -518,27 +518,35 @@ func (self *BoltStorage) StoreAuthSnapshot(
...
 		b := tx.Bucket([]byte(RATE_BUCKET))
 		c := b.Cursor()
-		_, lastEntry := c.Last()
-		if err := json.Unmarshal(lastEntry, &lastEntryjson); err != nil {
-			return err
+		lastKey, lastEntry := c.Last()
+		if lastKey == nil {
+			log.Printf("Bucket %s is empty", RATE_BUCKET)
+		} else {
+			if uErr = json.Unmarshal(lastEntry, &lastEntryJSON); uErr != nil {
```
cyberliem: the name is confuse. lastEntryJSON is common.AllRateEntry and lastEntry is []byte array represent the JSON string form of lastEntry?

data/storage/bolt.go
```diff
@@ -518,27 +518,35 @@ func (self *BoltStorage) StoreAuthSnapshot(
...
-	err = self.db.Update(func(tx *bolt.Tx) error {
-		var dataJson []byte
+	err := self.db.Update(func(tx *bolt.Tx) error {
+		var (
+			uErr          error
+			lastEntryJSON common.AllRateEntry
+			dataJson      []byte
+		)
+
 		b := tx.Bucket([]byte(RATE_BUCKET))
```
cyberliem: Can you double check if we should use CreateBucketIfNotExist so we won't get panic in case the db is new and hasn't been init with this bucket?

favadi: The bucket is created on initialization of the storage object https://github.com/KyberNetwork/reserve-data/blob/bbcf53d0b3a6fd1e1a839ae59bf49cb6b02abf09/data/storage/bolt.go#L85. We also have a card for review bucket creation everywhere: https://trello.com/c/yaVn6Fgl/83-createbucketifexists-in-db-operations so I change nothing here.

data/storage/bolt.go
favadi: I agree that it is confusing, I did the renaming in 3e49e60d18f1f72721a4bfc896b8baca9dc48fe6.

data/storage/bolt.go
cyberliem: Do we need to handle the case the bucket is empty ? Becasue tx.Bucket() can return empty.

cyberliem: Alright. Good to go now

PR #289, opened 2018-06-05
Add target qty v2 test

halink0803:


favadi:
[Write target quantity v2 test](https://trello.com/c/c1SudXQk/67-write-target-quantity-v2-test)

http/target_qty_v2_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+	"github.com/KyberNetwork/reserve-data/core"
+	"github.com/KyberNetwork/reserve-data/data"
+	"github.com/KyberNetwork/reserve-data/data/storage"
+	"github.com/KyberNetwork/reserve-data/http/httputil"
+	ethereum "github.com/ethereum/go-ethereum/common"
+	"github.com/gin-gonic/gin"
+)
+
+type newAssertFn func(t *testing.T, resp *httptest.ResponseRecorder)
```
favadi: what's different between this and `assertFn`.

http/target_qty_v2_test.go
```diff
@@ -0,0 +1,200 @@
...
+type targetQtytestCase struct {
+	msg      string
+	endpoint string
+	method   string
+	value    string
+	assert   newAssertFn
+}
```
favadi: this looks almost identical to `testCase`. Maybe instead of passing value/data we can accept a form data (of type map[string]string) instead?

http/target_qty_v2_test.go
```diff
@@ -0,0 +1,200 @@
...
+	t.Helper()
+
+	req, tErr := http.NewRequest(tc.method, tc.endpoint, nil)
+	if tErr != nil {
+		t.Fatal(tErr)
+	}
+
+	if tc.value != "" {
+		form := url.Values{}
+		form.Add("value", tc.value)
```
favadi: this looks almost identical to `testHTTPRequest` too, please reuse it.

PR #291, opened 2018-06-05
fix resource leaks in CallContract

favadi:


cyberliem:
[Resource leak in CallContract](https://trello.com/c/X6xokgiq/97-memory-leak-in-callcontract)

PR #296, opened 2018-06-06
Enable vet shadow

halink0803:


favadi:
[Avoid using shadowing variable](https://trello.com/c/7BcSvJ4f/66-avoid-using-shadowing-variable)

Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM golang
 ADD . /go/src/github.com/KyberNetwork/reserve-data
 
 ENV KYBER_EXCHANGES huobi,bittrex,binance
-WORKDIR /go/src/github.com/KyberNetwork/reserve-data
```
favadi: that could break production installation. What is the reason of this change?

blockchain/gas_oracle.go
```diff
@@ -42,8 +42,8 @@ func (self *GasOracle) RunGasPricing() error {
 		return err
 	}
 	defer func() {
-		if err := r.Body.Close(); err != nil {
-			log.Panic(err)
+		if vErr := r.Body.Close(); vErr != nil {
```
favadi: cErr.

Dockerfile
halink0803: If I change the config.json file to use relative path for keystore file. I have to change the Dockerfile as so to be able to run. If I don't change docker run break.

PR #308, opened 2018-06-11
Fix ignoring endpoind error and specifying message

halink0803:


exchange/binance/binance_endpoint.go
```diff
@@ -57,7 +57,10 @@ func (self *BinanceEndpoint) GetResponse(
 	client := &http.Client{
 		Timeout: time.Duration(30 * time.Second),
 	}
-	req, _ := http.NewRequest(method, url, nil)
+	req, newHTTPErr := http.NewRequest(method, url, nil)
```
favadi: in this case, you should move the err declaration in L72 to beginning of the function and reuse it.

favadi: I fixed it in eb66fd2c418207f993f085ccf305edab0a52b2ea.

PR #310, opened 2018-06-12
Fix init binance

halink0803:
Currently, binance is running with storage init by huobi: https://github.com/KyberNetwork/reserve-data/blob/4c4507869ac82502f1889bb834a1f87404cf71b9/cmd/configuration/exchanges.go#L97

which should be changed to use binance storage.


halink0803:
[binance exchange is using huobi storage instance](https://trello.com/c/CHJj8gq0/85-binance-exchange-is-using-huobi-storage-instance)

PR #314, opened 2018-06-18
fixed missing required fields in PWIs v2

favadi:


favadi:
[Authenticate PWIEquation with data field.](https://trello.com/c/ADPRMjDU/111-authenticate-pwiequation-with-data-field)

PR #318, opened 2018-06-19
refactor gas pricing

favadi:


favadi:
[GasOracle is not in used](https://trello.com/c/Pgonl1xC/114-gasoracle-is-not-in-used)

PR #320, opened 2018-06-20
Reorganize core setting

cyberliem:


favadi:
[Re organize all config in Core into one components, enable persistent storage for config](https://trello.com/c/esLGBVx6/12-re-organize-all-config-in-core-into-one-components-enable-persistent-storage-for-config)

PR #321, opened 2018-06-20
Add detail error message

halink0803:


halink0803:
[Return specific error message from Binance.](https://trello.com/c/cnTqWyV2/127-return-specific-error-message-from-binance)

exchange/binance/binance_endpoint.go
```diff
@@ -100,7 +100,16 @@ func (self *BinanceEndpoint) GetResponse(
...
 		break
 	default:
-		err = fmt.Errorf("Binance return with code: %d", resp.StatusCode)
+		respBody, err = ioutil.ReadAll(resp.Body)
+		if err != nil {
+			break
+		}
+		var response exchange.Binaresp
+		err = json.Unmarshal(respBody, &response)
+		if err != nil {
```
cyberliem: use if err= ..; err!=nil instead

exchange/binance/binance_endpoint.go
```diff
@@ -100,7 +100,15 @@ func (self *BinanceEndpoint) GetResponse(
...
 	default:
-		err = fmt.Errorf("Binance return with code: %d", resp.StatusCode)
+		respBody, err = ioutil.ReadAll(resp.Body)
+		if err != nil {
+			break
+		}
+		var response exchange.Binaresp
+		if err = json.Unmarshal(respBody, &response); err != nil {
+			break
+		}
```
favadi: all above lines can be replace with `json.NewDecoder(resp.Body).Decode`.

PR #324, opened 2018-06-20
Fix s3 upload path

cyberliem:
Fix Dirlog path error, update read me for aws config requirement


cyberliem:
[S3 archive doesn't work](https://trello.com/c/AYtiByh3/129-s3-archive-doesnt-work)

PR #325, opened 2018-06-21
Asserted interface conversion error. 

cyberliem:


cyberliem:
[Do not swallow interface conversion error in core](https://trello.com/c/E1j1WsEM/119-do-not-swallow-interface-conversion-error-in-core)

cmd/configuration/exchanges.go
```diff
@@ -147,15 +147,23 @@ func NewExchangePool(
 func (self *ExchangePool) FetcherExchanges() []fetcher.Exchange {
 	result := []fetcher.Exchange{}
 	for _, ex := range self.Exchanges {
-		result = append(result, ex.(fetcher.Exchange))
+		fcEx, ok := ex.(fetcher.Exchange)
+		if !ok {
+			log.Panic("Can not convert ExchangePool to fetcher exchange")
```
halink0803: It should be "Cannot" not "Can not"

cmd/configuration/exchanges.go
```diff
@@ -147,15 +147,23 @@ func NewExchangePool(
...
 	return result
 }
 
 func (self *ExchangePool) CoreExchanges() []common.Exchange {
 	result := []common.Exchange{}
 	for _, ex := range self.Exchanges {
-		result = append(result, ex.(common.Exchange))
+		cmEx, ok := ex.(common.Exchange)
+		if !ok {
+			log.Panic("Can not convert ExchangePool to core Exchange")
```
halink0803: "Cannot" not "Can not"

common/blockchain/broadcaster.go
```diff
@@ -40,7 +40,15 @@ func (self Broadcaster) Broadcast(tx *types.Transaction) (map[string]error, bool
 	wg.Wait()
 	result := map[string]error{}
 	failures.Range(func(key, value interface{}) bool {
-		result[key.(string)] = value.(error)
+		k, ok := key.(string)
```
halink0803: This is not really a handle because `func (key, value interface{}) bool` - whatever you return `true` or `false`, you don't handle it, so actually whatever you return it is useless.

core/reserve_core.go
```diff
@@ -52,8 +52,14 @@ func (self ReserveCore) CancelOrder(id common.ActivityID, exchange common.Exchan
 	if activity.Action != "trade" {
 		return errors.New("This is not an order activity so cannot cancel")
 	}
-	base := activity.Params["base"].(string)
-	quote := activity.Params["quote"].(string)
+	base, ok := activity.Params["base"].(string)
+	if !ok {
+		return fmt.Errorf("Can not convert params base (value: %v) to tokenID (type string)", activity.Params["base"])
```
halink0803: Should use "Cannot"

core/reserve_core.go
```diff
@@ -312,8 +318,24 @@ func (self ReserveCore) pendingSetrateInfo(minedNonce uint64) (*big.Int, *big.In
 	if act == nil {
 		return nil, nil, 0, nil
 	}
-	nonce, _ := strconv.ParseUint(act.Result["nonce"].(string), 10, 64)
-	gasPrice, _ := strconv.ParseUint(act.Result["gasPrice"].(string), 10, 64)
+	nonceStr, ok := act.Result["nonce"].(string)
```
halink0803: As it return error then you can return nil, nil, 0, err for easier return instead of init big.Int(0)

data/fetcher/fetcher.go
```diff
@@ -200,8 +200,18 @@ func (self *Fetcher) FetchAllAuthData(timepoint uint64) {
...
 		if found {
-			activityStatus := status.(common.ActivityStatus)
-			if activity.Result["tx"] != nil && activity.Result["tx"].(string) == "" {
+			activityStatus, ok := status.(common.ActivityStatus)
+			if !ok {
+				log.Print("ERROR: status from cexs can not be converted to common.ActivityStatus")
+				continue
+			}
+			//Set activity result tx to tx from cexs if currently result tx is not nil an is an empty string
+			resultTx, uOk := activity.Result["tx"].(string)
```
halink0803: Why using uOk instead of ok? It is not shadowing anything?

data/fetcher/fetcher.go
```diff
@@ -285,11 +295,16 @@ func (self *Fetcher) newNonceValidator() func(common.ActivityRecord) bool {
...
-		actNonce := act.Result["nonce"]
-		if actNonce == nil {
+		actNonce, ok := act.Result["nonce"].(string)
+		// interface assertion also return false if actNonce is nil
+		if !ok {
+			return false
+		}
+		nonce, err := strconv.ParseUint(actNonce, 10, 64)
+		if err != nil {
+			log.Printf("ERROR convert act.Result[nonce] to Uint64 failed %s", err)
```
halink0803: Should be err.Error() and also to be precise it should be act.Result[\"nonce\"]

data/fetcher/fetcher.go
```diff
@@ -303,7 +318,12 @@ func (self *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord)
 			var blockNum uint64
 			var status string
 			var err error
-			tx := ethereum.HexToHash(activity.Result["tx"].(string))
+			txStr, ok := activity.Result["tx"].(string)
+			if !ok {
+				log.Printf("ERROR: can not convert activity.Result[tx] (value %v) to string type", activity.Result["tx"])
```
halink0803: Should be cannot.

data/storage/bolt.go
```diff
@@ -653,6 +653,22 @@ func (self *BoltStorage) GetAllRecords(fromTime, toTime uint64) ([]common.Activi
...
+// interfaceConverstionToUint64 will assert the interface as string
+// and parse it to uint64. Return 0 if anything goes wrong)
+func interfaceConverstionToUint64(intf interface{}) uint64 {
+	numString, ok := intf.(string)
+	if !ok {
+		log.Printf("(%v) can't be converted to type string", intf)
+		return 0
+	}
+	num, uErr := strconv.ParseUint(numString, 10, 64)
+	if uErr != nil {
```
halink0803: Why use uErr instead of err, and it should be err.Error().

exchange/binance_format.go
```diff
@@ -12,8 +13,16 @@ type Binaprice struct {
 func (self *Binaprice) UnmarshalJSON(text []byte) error {
 	temp := []interface{}{}
 	err := json.Unmarshal(text, &temp)
```
halink0803: Should check this err != nil right away, because if err !=nil then interface conversion below will be failed.

stat/reserve_stat.go
```diff
@@ -218,7 +218,11 @@ func (self ReserveStats) GetHeatMap(fromTime, toTime uint64, tzparam int64) (com
 			return arrResult, err
 		}
 		for _, stat := range cStats {
-			s := stat.(common.MetricStats)
+			s, ok := stat.(common.MetricStats)
+			if !ok {
+				err = fmt.Errorf("Can not convert stat (%v) to MetricStat", s)
```
halink0803: "Cannot"

stat/reserve_stat.go
```diff
@@ -269,7 +273,11 @@ func (self ReserveStats) GetTokenHeatmap(fromTime, toTime uint64, tokenStr, freq
 			return arrResult, err
 		}
 		for _, stat := range stats {
-			s := stat.(common.VolumeStats)
+			s, ok := stat.(common.VolumeStats)
+			if !ok {
+				err = fmt.Errorf("Can not convert stat (%v) to VolumeStats", s)
+				continue
```
halink0803: Why continue when you return error != nil finally?. Return it right away will save operations below.

stat/reserve_stat.go
```diff
@@ -506,7 +514,11 @@ func (self ReserveStats) ExceedDailyLimit(address ethereum.Address) (bool, error
 				log.Printf("Got more than 1 day stats. This is a bug in GetUserVolume")
 			} else {
 				for _, volume := range volumeStats {
-					volumeValue := volume.(common.VolumeStats)
+					volumeValue, ok := volume.(common.VolumeStats)
+					if !ok {
+						log.Printf("Can not convert volume (%v) to VolumeStats", volume)
```
halink0803: "Cannot"

common/blockchain/broadcaster.go
cyberliem: according to Map documentation:
// Range calls f sequentially for each key and value present in the map.
// If f returns false, range stops the iteration.
So Range functions use this bool signal to decide if it should continue, not in the sense of error handling. In this sense, it should continue to check other records in case one record is malformed.

data/storage/bolt.go
cyberliem: shadowing err here, err is assigned to store the DB's transaction error.

stat/reserve_stat.go
cyberliem: as there are many records here, I think it's better to return the set of correct records and error to let the caller decided what to do.

data/fetcher/fetcher.go
halink0803: @cyberliem I think you miss this comment.

stat/reserve_stat.go
halink0803: The caller is us (we call it) and it will handle err != nil and send success = false with reason, so return error right away is more effective.

data/fetcher/fetcher.go
```diff
@@ -389,6 +409,61 @@ func unchanged(pre, post map[common.ActivityID]common.ActivityStatus) bool {
 	return true
 }
 
+func updateActivitywithBlockchainStatus(activity *common.ActivityRecord, bstatuses *sync.Map, snapshot *common.AuthDataSnapshot) {
+	status, _ := bstatuses.Load(activity.ID)
+	if status == nil {
```
favadi: please use the returned boolean value to check if the key exists.

data/fetcher/fetcher.go
```diff
@@ -389,6 +409,61 @@ func unchanged(pre, post map[common.ActivityID]common.ActivityStatus) bool {
...
+		activity.Result["status_error"] = activityStatus.Error.Error()
+	} else {
+		activity.Result["status_error"] = ""
+	}
+	activity.Result["blockNumber"] = activityStatus.BlockNumber
+}
+
+func updateActivitywithExchangeStatus(activity *common.ActivityRecord, estatuses *sync.Map, snapshot *common.AuthDataSnapshot) {
+	status, _ := estatuses.Load(activity.ID)
+	if status == nil {
```
favadi: please use the returned boolean value to check if the key exists.

exchange/bittrex.go
```diff
@@ -402,7 +413,18 @@ func (self *Bittrex) FetchTradeHistory() {
 			}
 			wait.Wait()
 			data.Range(func(key, value interface{}) bool {
-				result[key.(common.TokenPairID)] = value.([]common.TradeHistory)
+				tokenPairID, ok := key.(common.TokenPairID)
+				//if there is conversion error, continue to next key,val
```
favadi: is it OK to silently storeing incomplete trade history? Or it is better to return an error and try again next time? Please recheck other places with this same business logic.

cyberliem: After carefully rethink the matter, the logic is changed as follow: if the data returned from exchange is corrupted, skip this itter and wait till the next.

common/blockchain/broadcaster.go
favadi: Good point on this.

cyberliem:
[Unsafe type cast for common.Exchange](https://trello.com/c/I0aRzN6y/102-unsafe-type-cast-for-commonexchange)

cyberliem:
[Unsafe type cast for common.Exchange](https://trello.com/c/I0aRzN6y/102-unsafe-type-cast-for-commonexchange)

PR #326, opened 2018-06-22
fix typo: keytore to keystore

keyboard1989:


favadi:
@keyboard1989 thank you for your contribution!

PR #331, opened 2018-06-26
fix hardcoded API endpoints in string

favadi:


favadi:
[Do not hard code URL](https://trello.com/c/UFHXoyyj/87-do-not-hard-code-url)

halink0803:
@favadi LGTM, but I think you miss one case in cmd/clicmd/comparerates.go. Should it be hardcode or not?

favadi:
@halink0803 fixed in dfeeeb4f4835c01171f0536b26c946503080dc83.

PR #332, opened 2018-06-26
add and implement list-token API

cyberliem:


favadi:
[Add API to update core setting on the fly (list/delist token primarily)](https://trello.com/c/hzgUNqUN/126-add-api-to-update-core-setting-on-the-fly-list-delist-token-primarily)

http/settings_handler.go
```diff
@@ -44,48 +45,311 @@ func (self *HTTPServer) reloadTokenIndices(newToken common.Token, active bool) e
...
+			// If the pair is not in current token listing request, query it from exchange
+			_, ok := tokExSett.PrecisionLimit[pairID]
+			if !ok {
+				exchange, uErr := common.GetExchange(ex)
+				if uErr != nil {
+					httputil.ResponseFailure(c, httputil.WithReason(fmt.Sprintf("Can not get exchange instance %s (%s)", ex, uErr)))
+					return
+				}
+				epl, uErr := exchange.GetLiveExchangeInfo(pairID)
+				if (uErr != nil) && (!ok) {
```
favadi: the `!ok` is always true as we are in the scope `if !ok` (L281).

http/settings_handler.go
```diff
@@ -44,48 +45,311 @@ func (self *HTTPServer) reloadTokenIndices(newToken common.Token, active bool) e
...
+	if err := json.Unmarshal(data, &TokenListings); err != nil {
+		httputil.ResponseFailure(c, httputil.WithReason(fmt.Sprintf("cant not unmarshall token request %s", err.Error())))
+		return
+	}
+
+	pdPWI, err := self.metric.GetPendingPWIEquationV2()
+	if err != nil {
+		log.Printf("WARNING: can not get Pending PWIEquation v2, this will only allow listing non-internal token")
+	}
+	targetQty, err := self.metric.GetPendingTargetQtyV2()
```
favadi: The workflow here is not idea. To list a new token, user now need to call 3 HTTP APIs:

- create new PWIEquation
- create new TargetQty
- list Token

To list one new token, the user will have to reconfigured qty and pwi equation for **everything**. We should include the qty/pwi configuration in the token listing input too.

http/settings_handler.go
```diff
@@ -44,48 +45,311 @@ func (self *HTTPServer) reloadTokenIndices(newToken common.Token, active bool) e
...
 	if err != nil {
+		log.Printf("WARNING: can not get Pending TargetQty v2, this will only allow listing non-internal token")
+	}
+	// prepare each TokenListing instance for individual token
+	for tokenID, TokenListing := range TokenListings {
+		token := TokenListing.Token
+		// To list token, its active status must be true
+		if !token.Active {
```
favadi: So token.Active is now allowed to be `false` at all. Maybe we should not allow user to set it in first place?

http/settings_handler.go
```diff
@@ -44,48 +45,311 @@ func (self *HTTPServer) reloadTokenIndices(newToken common.Token, active bool) e
...
-	if internalBool {
-		if err = self.reloadTokenIndices(token, activeBool); err != nil {
+	httputil.ResponseSuccess(c, httputil.WithData(data))
+	return
+}
+
+// UpdateToken update minor independent details about a token
+// It provides the most simple way to modify  token's information without affecting other component
+// To list/delist token, use ListToken/ DelistToken API instead.
+func (self *HTTPServer) UpdateToken(c *gin.Context) {
```
favadi: As I review the implementation of this API, it will delete the stored token and replace with given token in post data. We should only update non-zero value of given token to the existing token setting stored in database.

settings/storage/token_bolt.go
```diff
@@ -209,3 +209,101 @@ func (boltSettingStorage *BoltSettingStorage) UpdateOneAddress(name settings.Add
...
+	err := boltSettingStorage.db.Update(func(tx *bolt.Tx) error {
+		//Apply tokens setting
+		for _, t := range tokens {
+			if uErr := addTokenByID(tx, t); uErr != nil {
+				return uErr
+			}
+			if uErr := addTokenByAddress(tx, t); uErr != nil {
+				return uErr
+			}
+		}
```
favadi: again, the name is Update but it is just override everything stored in database.

settings/storage/token_bolt.go
```diff
@@ -209,3 +209,101 @@ func (boltSettingStorage *BoltSettingStorage) UpdateOneAddress(name settings.Add
...
+}
+
+func (boltSettingStorage *BoltSettingStorage) StorePendingTokenListings(trs map[string]common.TokenListing) error {
+	err := boltSettingStorage.db.Update(func(tx *bolt.Tx) error {
+		b, uErr := tx.CreateBucketIfNotExists([]byte(PENDING_TOKEN_REQUEST))
+		if uErr != nil {
+			return uErr
+		}
+		for tokenID, tr := range trs {
+			dataJSON, vErr := json.Marshal(tr)
```
favadi: please declare dataJSON above and reuse the existing `uErr` error instance.

favadi: Also, do you know that uErr means "update error"?

settings/storage/token_bolt.go
```diff
@@ -209,3 +209,101 @@ func (boltSettingStorage *BoltSettingStorage) UpdateOneAddress(name settings.Add
...
+	result := make(map[string]common.TokenListing)
+	err := boltSettingStorage.db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(PENDING_TOKEN_REQUEST))
+		if b == nil {
+			return fmt.Errorf("Bucket %s does not exist yet", PENDING_TOKEN_REQUEST)
+		}
+		c := b.Cursor()
+		for k, v := c.First(); k != nil; k, v = c.Next() {
+			var tr common.TokenListing
+			if uErr := json.Unmarshal(v, &tr); uErr != nil {
```
favadi: do you know that vErr means error in view function?

settings/storage/token_bolt.go
```diff
@@ -209,3 +209,101 @@ func (boltSettingStorage *BoltSettingStorage) UpdateOneAddress(name settings.Add
...
+}
+
+func deletePendingTokenListings(tx *bolt.Tx) error {
+	b := tx.Bucket([]byte(PENDING_TOKEN_REQUEST))
+	if b == nil {
+		return fmt.Errorf("Bucket %s does not exist yet", PENDING_TOKEN_REQUEST)
+	}
+	c := b.Cursor()
+	for k, _ := c.First(); k != nil; k, _ = c.Next() {
+		if uErr := b.Delete(k); uErr != nil {
```
favadi: please name this error instance `err`.

http/settings_handler.go
```diff
@@ -44,48 +45,311 @@ func (self *HTTPServer) reloadTokenIndices(newToken common.Token, active bool) e
...
+		return nil, err
+	}
+	depAddrs, err := self.setting.GetDepositAddresses(exName)
+	if err != nil {
+		return nil, err
+	}
+	return common.NewCompositeExchangeSetting(depAddrs, exMinDep, exFee, exInfos), nil
+}
+
+func (self *HTTPServer) PrepareExchangeSetting(token common.Token, tokExSetts map[string]common.TokenExchangeSetting, preparedExchangeSetting *map[settings.ExchangeName]*common.CompositeExchangeSetting) error {
```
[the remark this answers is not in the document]
cyberliem: To ensure value change. However after I do a test with https://play.golang.org/p/xnk4mMTYUaZ apparently Map can be modified without passing pointer.

http/settings_handler.go
cyberliem: I'll discuss this with Analytic team and see if they prefer this solution. Personally I think currently the token Listing API is big enough already, while in fact we can shove PWIE, TargetQty (and QuadraticEQ as well) to TokenListing object, we'll have to handle the extra logic for this, while if we proceed with the current logic, we can be agnostic of other functionality of metrics, only need to reuse its interfaces.

http/settings_handler.go
cyberliem: The token struct contains a lot of field with primitive type (bool, float, int). To decide what is none-zero value of these types seems to be difficult. (for example, internal = false is the default value, but will it be non zero)?

We can alter these functions, accepts a range of params instead of a token Object individually  and update it into the currently available token. However the management for all these function is really messy. I recommend we keep this current implementation, and enforce end user to provide fully detailed token object.

settings/storage/token_bolt.go
cyberliem: As previously discussed, bolt DB's limitation leave us with update function to overwrite object with the same key. Throughout our implementation, we can only do actual "update" with list objects, which can be appends. I recommend we keep this implementation for now. 

I'll change function name to minimize confusion.

settings/storage/token_bolt.go
cyberliem: I know, yet I can't think of a name of inner-scope error

settings/storage/token_bolt.go
cyberliem: I know, yet I can't think of a name of inner-scope error. Pardon for my lack of creativity.

settings/storage/token_bolt.go
halink0803: as @favadi said, you should:
```
var dataJSON []byte
for tokenID, tr := range trs {
 dataJSON, uErr = json.Marshal(tr)
}
```

settings/storage/token_bolt.go
halink0803: You still can use vErr in this situation or err.

common/token.go
```diff
@@ -2,16 +2,38 @@ package common
...
+	Token       Token                           `json:"token"`
+	Exchange    map[string]TokenExchangeSetting `json:"exchanges"`
+	PWIEq       PWIEquationTokenV2              `json:"pwis_equation"`
+	TargetQty   TargetQtyV2                     `json:"target_qty"`
+	QuadraticEq RebalanceQuadraticEquation      `json:"rebalance_quadratic"`
+}
+
+type TokenFee struct {
+	Trading  float64
+	WithDraw float64
```
halink0803: Withdraw is a word and what is json parse for this type.

common/token.go
```diff
@@ -2,16 +2,38 @@ package common
...
+type TokenExchangeSetting struct {
+	DepositAddress string
+	Info           ExchangeInfo
+	Fee            TokenFee
+	MinDeposit     float64
+}
+
+type TokenListing struct {
+	Token       Token                           `json:"token"`
+	Exchange    map[string]TokenExchangeSetting `json:"exchanges"`
```
halink0803: This is one exchange or multiple exchanges? json parse is "exchanges" but name field does not have "s"?

common/token.go
```diff
@@ -2,16 +2,38 @@ package common
...
-	Decimal                 int64
-	Active                  bool
-	Internal                bool
-	MinimalRecordResolution string
-	MaxTotalImbalance       string
-	MaxPerBlockImbalance    string
+	ID                      string `json:"id"`
+	Name                    string `json:"name"`
+	Address                 string `json:"address"`
+	Decimal                 int64  `json:"decimals"`
```
halink0803: "Decimal" or "Decimals", why json parse is different?

data/storage/bolt.go
```diff
@@ -1689,3 +1690,109 @@ func (self *BoltStorage) GetRebalanceQuadratic() (metric.RebalanceQuadraticReque
 	})
 	return result, err
 }
+
+func (self *BoltStorage) StoreJSONByteArray(tx *bolt.Tx, bucketName string, key, value []byte) error {
```
halink0803: Please add comment for this function usage. This function name general but it seems this function only for store pending bucket?

data/storage/bolt.go
```diff
@@ -1689,3 +1690,109 @@ func (self *BoltStorage) GetRebalanceQuadratic() (metric.RebalanceQuadraticReque
...
+	c := b.Cursor()
+	_, v := c.First()
+	if v != nil {
+		return fmt.Errorf("Bucket %s has a pending record", bucketName)
+	}
+	return b.Put(key, value)
+}
+
+func (self *BoltStorage) StorePendingTokenListingInfo(tarQty common.TokenTargetQtyV2, pwi common.PWIEquationRequestV2, quadEq common.RebalanceQuadraticRequest) error {
+	timeStampkey := boltutil.Uint64ToBytes(common.GetTimepoint())
```
halink0803: Naming convention should be timeStampKey or timestampKey, not timeStampkey.

data/storage/bolt.go
```diff
@@ -1689,3 +1690,109 @@ func (self *BoltStorage) GetRebalanceQuadratic() (metric.RebalanceQuadraticReque
...
+	err := self.db.Update(func(tx *bolt.Tx) error {
+		dataJSON, vErr := json.Marshal(tarQty)
```
halink0803: Do you know uErr stand for "update error"? Please change it.

halink0803: You can use `uErr` for the whole update function. Below you should use `if dataJSON, uErr = ...; uErr != nil {}`

data/storage/bolt.go
halink0803: Does this function exported? If not, please make it private.

data/storage/bolt.go
```diff
@@ -1689,3 +1690,109 @@ func (self *BoltStorage) GetRebalanceQuadratic() (metric.RebalanceQuadraticReque
...
+		}
+		return nil
+	})
+	return err
+}
+
+func (self *BoltStorage) ConfirmTokenListingInfo(tarQty common.TokenTargetQtyV2, pwi common.PWIEquationRequestV2, quadEq common.RebalanceQuadraticRequest) error {
+	timeStampkey := boltutil.Uint64ToBytes(common.GetTimepoint())
+	err := self.db.Update(func(tx *bolt.Tx) error {
+		dataJSON, vErr := json.Marshal(tarQty)
```
halink0803: Use uErr only inside this function.

exchange/binance.go
```diff
@@ -139,7 +160,11 @@ func (self *Binance) UpdatePairsPrecision() error {
 		return errors.New("Exchange info of Binance is nil")
 	}
 	for pair := range exInfo.GetData() {
-		self.UpdatePrecisionLimit(pair, symbols, &exInfo)
+		epl, exist := self.getPrecisionLimitFromSymbols(pair, symbols)
```
halink0803: This ```epl``` variable is not very clear. Please use better name, don't mind of long variable.

exchange/bittrex.go
```diff
@@ -123,11 +132,27 @@ func (self *Bittrex) UpdatePairsPrecision() error {
 		return errors.New("Exchange info of Bittrex is nil")
 	}
 	for pair := range exInfo.GetData() {
-		self.UpdatePrecisionLimit(pair, symbols, &exInfo)
+		epl, exist := self.getPrecisionLimitFromSymbols(pair, symbols)
```
halink0803: Same here.

exchange/huobi.go
```diff
@@ -118,7 +138,11 @@ func (self *Huobi) UpdatePairsPrecision() error {
 		return errors.New("Exchange info of Huobi is nil")
 	}
 	for pair := range exInfo.GetData() {
-		self.UpdatePrecisionLimit(pair, exchangeInfo, &exInfo)
+		epl, exist := self.getPrecisionLimitFromSymbols(pair, exchangeInfo)
```
halink0803: Same here.

exchange/stable_ex.go
```diff
@@ -48,6 +48,10 @@ func (self *StableEx) GetInfo() (common.ExchangeInfo, error) {
 	return self.setting.GetExchangeInfo(settings.StableExchange)
 }
 
+func (self *StableEx) GetLiveExchangeInfos(tokenPairIDs []common.TokenPairID) (common.ExchangeInfo, error) {
```
halink0803: Naming should be GetLiveExchangeInfo (info has no "s")

settings/token.go
```diff
@@ -86,3 +86,19 @@ func (setting *Settings) MustCreateTokenPair(base, quote string) common.TokenPai
 func (setting *Settings) UpdateToken(t common.Token) error {
 	return setting.Tokens.Storage.UpdateToken(t)
 }
+
+func (setting *Settings) ApplyTokenWithExchangeSetting(tokens []common.Token, exSetting map[ExchangeName]*common.ExchangeSetting) error {
+	return setting.Tokens.Storage.UpdateTokenWithExchangeSetting(tokens, exSetting)
+}
+
+func (setting *Settings) UpdatePendingTokenListings(trs map[string]common.TokenListing) error {
```
halink0803: what does ```trs``` mean?

settings/storage/token_bolt.go
halink0803: It seems you did not change the function name?

settings/storage/exchange_bolt.go
```diff
@@ -35,19 +35,23 @@ func (boltSettingStorage *BoltSettingStorage) GetFee(ex settings.ExchangeName) (
...
-			return uErr
-		}
-		return b.Put(boltutil.Uint64ToBytes(uint64(ex)), dataJSON)
+		return putFee(tx, ex, data)
 	})
 	return err
 }
 
+func putFee(tx *bolt.Tx, ex settings.ExchangeName, fee common.ExchangeFees) error {
+	b, uErr := tx.CreateBucketIfNotExists([]byte(EXCHANGE_FEE_BUCKET))
```
halink0803: This bucket should be initiate from ```NewStorage``` function, should not CreateBucketIfNotExists here because it is not a nested bucket.

halink0803: apply to all ```put``` functions.

settings/storage/address_bolt.go
```diff
@@ -10,6 +10,53 @@ import (
...
+			keyName := settings.AddressName(boltutil.BytesToUint64(k)).String()
+			result[keyName] = string(v)
+		}
+		b = tx.Bucket([]byte(ADDRESS_SET_SETTING_BUCKET))
+		if b == nil {
+			return fmt.Errorf("Bucket %s has not existed yet", ADDRESS_SET_SETTING_BUCKET)
+		}
+		c = b.Cursor()
+		for k, _ := c.First(); k != nil; k, _ = c.Next() {
+			nestedBuck := b.Bucket(k)
```
halink0803: Why not checking for bucket existent here? plus, for nested bucket should use ```CreateBucketIfNotExist```.

README.md
```diff
@@ -1571,6 +1571,452 @@ response
...
+response
+
+```
+on success:
+{"success":true}
+on failure:
+{"success":false,
+ "reason":<error>}
+```
+
```
halink0803: It seems you do not add enough readme for all new apis you wrote. Please add them to readme.

common/token.go
cyberliem: Updated this struct with json tag following our fee.json

common/token.go
cyberliem: Don't really know, in our json config file it was decimals, in core implementation from beginning until now it is decimal. Please advise if there is really a need to change any of them.

data/storage/bolt.go
cyberliem: This is value error. Not update error.

data/storage/bolt.go
cyberliem: This is value error. Not update error.

exchange/stable_ex.go
cyberliem: The "s" implied that this function get multiple instance of exchanges info, to distinguish with the other function which get only one instance of this.

settings/storage/address_bolt.go
cyberliem: View Tx can not create bucket. As mentioned earlier, there is a card for this approach already. Making change here is out of scope.

README.md
```diff
@@ -1570,7 +1570,745 @@ response
     "success": true,
   }
 ```
+# Setting APIs
```
favadi: pleas use correct header level.

README.md
```diff
@@ -1570,7 +1570,745 @@ response
...
+## Token related APIs
+### Update token - (signing required)
```
favadi: please update the document to specified that /setting/update-token will replace the stored token with given one in input data, not update it.

README.md
```diff
@@ -1570,7 +1570,745 @@ response
...
+      },
+      \"NEO\": {
+        \"Token\": {
+          \"id\": \"NEO\",
+          \"name\": \"Request\",
+          \"decimals\": 18,
+          \"address\": \"0x8f8221afbb33998d8584a2b05749ba73c37a938a\",
+          \"minimalRecordResolution\": \"1000000000000000\",
+          \"maxPerBlockImbalance\": \"27470469074054960644096\",
+          \"maxTotalImbalance\": \"33088179999699195920384\",
```
favadi: they should be minimal_record_resolution, ...?

README.md
```diff
@@ -1570,7 +1570,745 @@ response
...
+{data : {
+  "AST": {
+        "Token": {
+          "id": "AST",
+          "name": "AirSwap",
+          "address": "0x27054b13b1b798b345b591a4d22e6562d47ea75a",
+          "decimals": 18,
+          "active": True,
+          "internal": False,
+          "minimalRecordResolution": "10",
```
favadi: again, they should be minimal_record_resolution, ...?

common/token.go
```diff
@@ -2,16 +2,37 @@ package common
...
+	DepositAddress string
+	Info           ExchangeInfo
+	Fee            TokenFee
+	MinDeposit     float64
+}
+
+type TokenListing struct {
+	Token       Token                           `json:"token"`
+	Exchanges   map[string]TokenExchangeSetting `json:"exchanges"`
+	PWIEq       PWIEquationTokenV2              `json:"pwis_equation"`
```
favadi: Maybe the tasks should be pwi_equations?

common/token.go
favadi: Just call the field `Decimals`?

common/metric_types.go
```diff
@@ -1,4 +1,4 @@
-package metric
+package common
```
favadi: what is the reason of this renaming?

data/storage/bolt.go
```diff
@@ -1689,3 +1690,106 @@ func (self *BoltStorage) GetRebalanceQuadratic() (metric.RebalanceQuadraticReque
...
 }
+
+func (self *BoltStorage) storeJSONByteArray(tx *bolt.Tx, bucketName string, key, value []byte) error {
+	b := tx.Bucket([]byte(bucketName))
+	if b == nil {
+		return fmt.Errorf("Bucket %s hasn't existed yet", bucketName)
+	}
+	c := b.Cursor()
+	_, v := c.First()
+	if v != nil {
```
favadi: checking the key if it != nil should be better.

data/storage/bolt.go
```diff
@@ -1689,3 +1690,106 @@ func (self *BoltStorage) GetRebalanceQuadratic() (metric.RebalanceQuadraticReque
 	})
 	return result, err
 }
+
+func (self *BoltStorage) storeJSONByteArray(tx *bolt.Tx, bucketName string, key, value []byte) error {
```
favadi: please document that this function will only insert the record to bucket only if it is empty.

data/storage/bolt.go
```diff
@@ -1689,3 +1690,106 @@ func (self *BoltStorage) GetRebalanceQuadratic() (metric.RebalanceQuadraticReque
...
+	err := self.db.Update(func(tx *bolt.Tx) error {
+		dataJSON, uErr := json.Marshal(tarQty)
+		if uErr != nil {
+			return uErr
+		}
+		if uErr = self.storeJSONByteArray(tx, PENDING_TARGET_QUANTITY_V2, []byte("current_pending_target_qty"), dataJSON); uErr != nil {
+			return uErr
+		}
+		dataJSON, uErr = json.Marshal(pwi)
+		if uErr != nil {
```
favadi: please join 1717-1718

data/storage/bolt.go
```diff
@@ -1689,3 +1690,106 @@ func (self *BoltStorage) GetRebalanceQuadratic() (metric.RebalanceQuadraticReque
...
+			return uErr
+		}
+		if uErr = self.storeJSONByteArray(tx, PENDING_PWI_EQUATION_V2, timeStampKey, dataJSON); uErr != nil {
+			return uErr
+		}
+		dataJSON, uErr = json.Marshal(quadEq)
+		if uErr != nil {
```
favadi: please join 1724-1725

common/metric_types.go
cyberliem: Metric type is now moved to common, since to integrate between metric ops (PWIS, quadratic, TargetQuantity) and token listing ops, these two types must be composited into one struct storing both these type. If the metric type is still at metric package, it will induce cycle importing.

common/token.go
cyberliem: Changed, but a month from now when this branch merge into core, it will case a hell of conflict while the old core is using `Decimal`, while new implementation use `Decimals`

PR #333, opened 2018-06-26
Hot fix snapshot valid

halink0803:


data/fetcher/fetcher.go
```diff
@@ -415,13 +415,15 @@ func (self *Fetcher) PersistSnapshot(
...
 					newEbalance := oldAuth.ExchangeBalances[key.(common.ExchangeID)]
+					newEbalance.Valid = false
 					newEbalance.Error = v.Error
 					newEbalance.Status = false
 					allEBalances[key.(common.ExchangeID)] = newEbalance
 				}
 			}
-			snapshot.Valid = false
-			snapshot.Error = v.Error
+			// Temporarily snapshot is false only when cannot get balance from blockchain
```
cyberliem: should it be invalid if newEbalance.Valid =false and newEbanalnce.Status= true?

halink0803:
Binance not maintainnace anymore. This fix should be applied to next staging.

PR #334, opened 2018-06-26
Validate exchange info

halink0803:


halink0803:
[Validate exchange info before return through api](https://trello.com/c/ZOTzZjTz/140-validate-exchange-info-before-return-through-api)

cmd/configuration/getconfig.go
```diff
@@ -56,7 +56,7 @@ func GetConfigPaths(kyberENV string) SettingPaths {
 func GetConfig(kyberENV string, authEnbl bool, endpointOW string, noCore, enableStat bool) *Config {
 	setPath := GetConfigPaths(kyberENV)
 
-	world, err := world.NewTheWorld(kyberENV, setPath.secretPath)
+	newWorld, err := world.NewTheWorld(kyberENV, setPath.secretPath)
```
favadi: it should be named `theWorld` instead, as the constructor is called `NewTheWorld`.

common/types.go
```diff
@@ -107,27 +107,30 @@ func (self *ExchangeInfo) Update(pair TokenPairID, data ExchangePrecisionLimit)
...
+//Get return Exchange precision limit of on token pair of an exchange
 func (self *ExchangeInfo) Get(pair TokenPairID) (ExchangePrecisionLimit, error) {
 	self.mu.RLock()
 	defer self.mu.RUnlock()
 	if info, exist := self.data[pair]; exist {
 		return info, nil
-	} else {
-		return info, fmt.Errorf("Token pair is not existed")
 	}
+	return ExchangePrecisionLimit{}, fmt.Errorf("Token pair is not existed")
```
favadi: please use `errors.New` here.

common/exchange.go
```diff
@@ -10,6 +10,7 @@ import (
 // Exchange represents a centralized exchange like Binance, Huobi...
 type Exchange interface {
 	ID() ExchangeID
+	Pairs() []TokenPair
```
favadi: Could you add a comment doc to his new method?

exchange/binance.go
```diff
@@ -109,10 +109,11 @@ func (self *Binance) UpdatePrecisionLimit(pair common.TokenPair, symbols []Binan
 	}
 }
 
+//UpdatePairsPrecision update precision for trading for pairs
 func (self *Binance) UpdatePairsPrecision() {
 	exchangeInfo, err := self.interf.GetExchangeInfo()
 	if err != nil {
-		log.Printf("Get exchange info failed: %s\n", err)
+		log.Panicf("Get exchange info from binance failed: %s", err.Error())
```
favadi: Maybe this function should return an error instead?

halink0803: This function can return error. However, this function runs when core started and it will stop the core if cannot get info for its supported tokens from exchanges.

favadi: Great, please change it to return an error, the caller will decide if it want to panic or not.

PR #335, opened 2018-06-26
update staging config with new tokens

favadi:


favadi:
[Update fee and huobi config for new token listing](https://trello.com/c/eD4fQjWq/144-update-fee-and-huobi-config-for-new-token-listing)

PR #336, opened 2018-06-27
Update valid function for pwi

halink0803:


halink0803:
[IsValid() function should return detail error.](https://trello.com/c/pHhZIJVI/149-isvalid-function-should-return-detail-error)

PR #337, opened 2018-06-28
Hot fix trade history qty is not saved

halink0803:


halink0803:
[CEX trade list returns quantity 0](https://trello.com/c/5bh9PeVs/152-cex-trade-list-returns-quantity-0)

PR #340, opened 2018-06-28
Exclude stable exchange from exchange info validation

halink0803:


halink0803:
@favadi I updated the comment for excluding stable exchange from checking for supported tokens.

favadi:
[Validate exchange info before return through api](https://trello.com/c/ZOTzZjTz/140-validate-exchange-info-before-return-through-api)

PR #341, opened 2018-06-29
add sort before storing logs

cyberliem:
To fix the potential skipping logs if store log fail, the logs must be sorted in increasing order of blockNumber before processing


cyberliem:
[Review logic of log fetcher in stat/fetcher.go.](https://trello.com/c/2HleHStn/71-review-logic-of-log-fetcher-in-stat-fetchergo)

PR #342, opened 2018-06-29
changed logic flow of buildTx

cyberliem:


cyberliem:
[In successive blockchain.send(), if one transaction fail, the next transactions with a higher nonce will lost, which will stall all the other send() operation.](https://trello.com/c/qjUjOq4j/157-in-successive-blockchainsend-if-one-transaction-fail-the-next-transactions-with-a-higher-nonce-will-lost-which-will-stall-all-th)

PR #343, opened 2018-06-29
make pr_diff.sh allow timestamp querry

favadi:


favadi:
[Monthly core development progress report (2018-June)](https://trello.com/c/skIKckJe/160-monthly-core-development-progress-report-2018-june)

pr_diff.sh
```diff
@@ -1,17 +1,49 @@
 #!/bin/bash
```
vietthang207: What is the purpose of this script? From my understanding, it just add an url of the pull request?

pr_diff.sh
```diff
@@ -1,17 +1,49 @@
...
+	;;
+    "*")
+	print_usage
+	exit 1
+esac
 
 for commit_id in "${pr_commit_ids[@]}"; do
     git show --format="%B" "$commit_id" |\
-	awk -v github_url="$github_url" 'NR == 1 {$4="["$4"]("github_url substr($4,2)")"; print "* "$0; next}; {print "  "$0}'
+	awk -v github_url="$github_url" 'NR == 1 {$4="["$4"]("github_url substr($4,2)")"; print "* "$0; next}; /^diff --/{exit}; {print "  "$0}'
```
vietthang207: I seems to me that "git show" will not show any diff message, do you have an example of such an output?

vietthang207: I've try "git show -m commit_ids", it will show some diff if the commit is a merge commit.

favadi: > I seems to me that "git show" will not show any diff message, do you have an example of such an output?

You could just run git show in any of your repos.

pr_diff.sh
favadi: To generate a changelog from git history.

PR #344, opened 2018-06-29
Fix error trade history ignore not supported token

halink0803:


halink0803:
[Fix error: token no longer supported, then core does not get trade history for that token even though it should.](https://trello.com/c/2gZ7Arqr/161-fix-error-token-no-longer-supported-then-core-does-not-get-trade-history-for-that-token-even-though-it-should)

PR #345, opened 2018-07-02
Update persist auth snapshot

halink0803:


halink0803:
[Return error if get blockchain balance error.](https://trello.com/c/O6AcVMni/151-return-error-if-get-blockchain-balance-error)

data/fetcher/fetcher.go
```diff
@@ -529,6 +529,18 @@ func (self *Fetcher) PersistSnapshot(
 	}
 	// note: only update status when it's pending status
 	snapshot.ExchangeBalances = allEBalances
+
+	// persist blockchain balance
+	// if blockchain balance is not valid then auth snapshot will also not valid
+	for _, balance := range bbalances {
+		if !balance.Valid {
+			snapshot.Valid = false
+			if balance.Error != "" {
```
favadi: so, what happens if balance.Error is not empty? the error is ignored?

data/fetcher/fetcher.go
```diff
@@ -529,6 +529,18 @@ func (self *Fetcher) PersistSnapshot(
...
+				snapshot.Error += " " + balance.Error
```
favadi: Could you add a `;` between error messages?

data/fetcher/fetcher.go
halink0803: I thought it would be non-sense to add an empty string. I added the error message.

PR #358, opened 2018-07-12
Update dev settings and huobi endpoint error handler

halink0803:


halink0803:
[Update Huobi signature signing scheme](https://trello.com/c/32JsGVqV/181-update-huobi-signature-signing-scheme)

PR #361, opened 2018-07-17
Modify token listing

cyberliem:


cyberliem:
[modify token listing API for analytic change](https://trello.com/c/Fd5yGuMi/184-modify-token-listing-api-for-analytic-change)

http/settings_handler.go
```diff
@@ -472,6 +440,13 @@ func (self *HTTPServer) UpdateToken(c *gin.Context) {
 			return
 		}
 	}
+	currTok, err := self.setting.GetTokenByID(token.ID)
+	if (err != nil) || ((err == nil) && (currTok.Active != token.Active)) {
```
favadi: please define a `TokenNotFound` error in storage interface, and catch only the error happens when that happen to handle here. For other error instances, just return to client.

PR #365, opened 2018-07-19
refactor set rate

cyberliem:


cyberliem:
[Refactor logic flow in SetRates function](https://trello.com/c/agGT4IdL/172-refactor-logic-flow-in-setrates-function)

core/reserve_core.go
```diff
@@ -339,6 +339,65 @@ func (self ReserveCore) pendingSetrateInfo(minedNonce uint64) (*big.Int, *big.In
 	return big.NewInt(int64(nonce)), big.NewInt(int64(gasPrice)), count, nil
 }
 
+func (self ReserveCore) GetSetRateResult(tokens []common.Token,
+	buys []*big.Int,
+	sells []*big.Int,
+	block *big.Int,
+	afpMids []*big.Int) (error, *types.Transaction) {
```
favadi: the function signature could be changed a bit so we can write:

```go
func (self ReserveCore) GetSetRateResult(tokens []common.Token,
	buys, sells, afpMids []*big.Int,
	block *big.Int) (error, *types.Transaction) {
```

favadi: Actually, the error should be second value in return, please change it to.

```go
func (self ReserveCore) GetSetRateResult(tokens []common.Token,
	buys, sells, afpMids []*big.Int,
	block *big.Int) (*types.Transaction, error) {
```

core/reserve_core.go
```diff
@@ -339,6 +339,65 @@ func (self ReserveCore) pendingSetrateInfo(minedNonce uint64) (*big.Int, *big.In
...
+	var (
+		tx  *types.Transaction
+		err error
+	)
+	if len(tokens) != len(buys) || len(tokens) != len(sells) || len(tokens) != len(afpMids) {
+		return errors.New("Tokens, buys sells and afpMids must have the same length"), tx
+	}
```
favadi: I know it is repetitive but please change this to something like:

```go
if len(tokens) != len(buys) {
return fmt.Errorf("number of buys (%d) is not equal to number of tokens (%d), len(buys), len(tokens)
```

This way user will have a better error message.

core/reserve_core.go
```diff
@@ -339,6 +339,65 @@ func (self ReserveCore) pendingSetrateInfo(minedNonce uint64) (*big.Int, *big.In
...
+
+	if err = sanityCheck(buys, afpMids, sells); err != nil {
+		return err, tx
+	}
+	tokenAddrs := []ethereum.Address{}
```
favadi: We should use `var tokenAddrs []ethereum.Address` instead. Reference: https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices.

core/reserve_core.go
```diff
@@ -339,6 +339,65 @@ func (self ReserveCore) pendingSetrateInfo(minedNonce uint64) (*big.Int, *big.In
...
+	for _, token := range tokens {
+		tokenAddrs = append(tokenAddrs, ethereum.HexToAddress(token.Address))
+	}
+	// if there is a pending set rate tx, we replace it
+	var oldNonce *big.Int
+	var oldPrice *big.Int
+	var minedNonce uint64
+	var count uint64
```
favadi: those should be grouped  in form

```go
var (
xxx
yyy
)
```

PR #368, opened 2018-07-23
Update user aggregation with token-token eth amount

halink0803:


halink0803:
[Wrong total_eth_volume value of /get-user-list API](https://trello.com/c/P1UuGn5E/186-wrong-totalethvolume-value-of-get-user-list-api)

PR #372, opened 2018-07-23
add and applied universal constants

cyberliem:


cyberliem:
[Use constants for exchange/ mining status and activities strings.](https://trello.com/c/3mjmaqDd/187-use-constants-for-exchange-mining-status-and-activities-strings)

PR #375, opened 2018-07-24
change log message to WARNING instead of ERROR

cyberliem:


cyberliem:
[Assertion error in core failed the Simulation.](https://trello.com/c/C7yZciOf/191-assertion-error-in-core-failed-the-simulation)

cyberliem:
[Assertion error in core failed the Simulation.](https://trello.com/c/C7yZciOf/191-assertion-error-in-core-failed-the-simulation)

cyberliem:
[IP locator should be init from server start, instead of calling it in every IPtoCountry function.](https://trello.com/c/DtQ7gZVJ/192-ip-locator-should-be-init-from-server-start-instead-of-calling-it-in-every-iptocountry-function)

PR #376, opened 2018-07-24
reviewed and modify the unnecessary panics

cyberliem:


cyberliem:
[Review all panic usages](https://trello.com/c/GyYAeybZ/105-review-all-panic-usages)

common/blockchain/base_blockchain.go
```diff
@@ -66,16 +67,22 @@ func (self *BaseBlockchain) RecommendedGasPriceFromNode() (*big.Int, error) {
 	return self.client.SuggestGasPrice(timeout)
 }
 
+// GetOperator returns the operator if avail, return nil if the operator can't be found
 func (self *BaseBlockchain) GetOperator(name string) *Operator {
 	op, found := self.operators[name]
 	if !found {
-		panic(fmt.Sprintf("operator %s is not found. you have to register it before using it", name))
+		log.Printf("operator %s is not found. you have to register it before using it", name)
```
halink0803: This must be panic here. Cannot change.

common/blockchain/base_blockchain.go
```diff
@@ -97,9 +108,13 @@ func (self *BaseBlockchain) GetNextNonce(operator string) (*big.Int, error) {
...
 func (self *BaseBlockchain) SignAndBroadcast(tx *types.Transaction, from string) (*types.Transaction, error) {
-	signer := self.GetOperator(from).Signer
+	op := self.GetOperator(from)
+	if op == nil {
+		return nil, fmt.Errorf("cannot find operator %s", from)
+	}
+	signer := op.Signer
 	if tx == nil {
-		panic(errors.New("Nil tx is forbidden here"))
+		return nil, errors.New("Nil tx is forbidden here")
```
halink0803: Here also, must be panic because cannot return nil tx.

common/blockchain/transaction.go
```diff
@@ -73,8 +73,10 @@ func (s *senderFromServer) Sender(tx *types.Transaction) (ethereum.Address, erro
 }
 
 func (s *senderFromServer) Hash(tx *types.Transaction) ethereum.Hash {
+	// This should never happen
```
halink0803: This is an exported function, you can comment above function name, increase golint point and explain also.

common/blockchain/transaction.go
```diff
@@ -73,8 +73,10 @@ func (s *senderFromServer) Sender(tx *types.Transaction) (ethereum.Address, erro
...
 	panic("can't sign with senderFromServer")
 }
 func (s *senderFromServer) SignatureValues(tx *types.Transaction, sig []byte) (R, S, V *big.Int, err error) {
+	// This should never happen
```
halink0803: Here also.

common/types.go
```diff
@@ -18,6 +18,7 @@ type Timestamp string
 
 func (self Timestamp) ToUint64() uint64 {
 	res, err := strconv.ParseUint(string(self), 10, 64)
+	//  this should never happen. Timestamp is never manually entered.
```
halink0803: It can happen if Timestamp is not correctly format.

exchange/binance.go
```diff
@@ -572,7 +572,7 @@ func (self *Binance) WithdrawStatus(id, currency string, amount float64, timepoi
 func (self *Binance) OrderStatus(id string, base, quote string) (string, error) {
 	orderID, err := strconv.ParseUint(id, 10, 64)
 	if err != nil {
-		panic(err)
+		return "", fmt.Errorf("Can not parse orderID (val %v) to uint", id)
```
halink0803: Can print id with %s.

exchange/stable_ex.go
```diff
@@ -36,12 +36,8 @@ func (self *StableEx) Address(token common.Token) (ethereum.Address, bool) {
 	return addr, supported
 }
 
-func (self *StableEx) UpdateAllDepositAddresses(address string) {
-	panic("dgx doesn't support update deposit addresses")
-}
-
 func (self *StableEx) UpdateDepositAddress(token common.Token, address string) error {
-	panic("dgx doesn't support update deposit addresses")
+	return errors.New("dgx doesn't support update deposit addresses")
```
halink0803: Victor panic here to make sure UpdateDepositAddress will never be called in code. Can return error but panic or fatal is better (we cannot ignore it)

settings/storage/setting_bolt.go
```diff
@@ -89,7 +89,7 @@ func NewBoltSettingStorage(dbPath string) (*BoltSettingStorage, error) {
 		return nil
 	})
 	if err != nil {
-		panic(err)
+		return nil, err
```
halink0803: Cannot init db seem crucial. Are you sure you will not miss this error?

stat/storage/analytic_bolt.go
```diff
@@ -30,7 +30,7 @@ func NewBoltAnalyticStorage(dbPath string) (*BoltAnalyticStorage, error) {
 	var db *bolt.DB
 	db, err = bolt.Open(dbPath, 0600, nil)
 	if err != nil {
-		panic(err)
+		return nil, err
```
halink0803: Stat package is not very important in reserve data scope, but separatedly this "cannot init database" should be panic.

common/blockchain/base_blockchain.go
cyberliem: The later code propose different: possible to return nil and error. 

if signedTx != nil {
				return signedTx, fmt.Errorf("Broadcasting transaction %s failed, retry failures: %s", tx.Hash().Hex(), failures)
			} else {
				return signedTx, fmt.Errorf("Broadcasting transaction failed, retry failures: %s", failures)
			}

common/types.go
cyberliem: Where in the code can Timestamp be input with incorrect format?

exchange/stable_ex.go
cyberliem: If someone call UpdateDepositAddress by mistake error will be return and nothing change. It doesn't stop the core functionality or to affect the execution integrity in anyway. If you think there could be problem associate with this, please provide example.

settings/storage/setting_bolt.go
cyberliem: The caller of this DB object will decide if it can continue without the db. And yes, setting db init error will panic at start server.

stat/storage/analytic_bolt.go
cyberliem: same as above.

common/types.go
halink0803: Anywhere call ToUint64(), to be general we don't care where call it, we only know that Timestamp only a string value, and it can be incorrect (space at ends, or between numbers), then err "can" happen.

exchange/stable_ex.go
halink0803: Yeah, I think it might bring problem in the future. Panic here make sure no ones call it by mistake. return errors does not improve anything anyway.

common/blockchain/base_blockchain.go
```diff
@@ -222,6 +224,9 @@ func (self *BaseBlockchain) GetCallOpts(block uint64) CallOpts {
 func (self *BaseBlockchain) GetTxOpts(op string, nonce *big.Int, gasPrice *big.Int, value *big.Int) (TxOpts, error) {
 	result := TxOpts{}
 	operator := self.GetOperator(op)
+	if operator == nil {
```
halink0803: Those code will never be reached.

exchange/stable_ex.go
cyberliem: So there is no scenario where this can be problematic? As explained:
- This can happen (calling into exchange.UpdateDepositAddress() without aware which exchange is calling). It's not "should not happen case", hence panic use is not a must.
- When this happen, nothing change. This contradict to the reason to panic, i.e, program can not function anymore.   

Hence the panic is removed. Unless provided with concrete case, I found no reason to keep panic here.

common/types.go
cyberliem: Our implementation of core is to never manually enter timestamp but to call common.GetTimeStamp(). You can verify it if necessary.

common/blockchain/base_blockchain.go
```diff
@@ -54,6 +54,7 @@ func (self *BaseBlockchain) OperatorAddresses() map[string]ethereum.Address {
 }
 
 func (self *BaseBlockchain) RegisterOperator(name string, op *Operator) {
+	//This shouldn't happen, each operator get registered only once.
```
favadi: this function should be renamed to MustRegisterOperator then.

common/blockchain/base_blockchain.go
```diff
@@ -66,6 +67,7 @@ func (self *BaseBlockchain) RecommendedGasPriceFromNode() (*big.Int, error) {
 	return self.client.SuggestGasPrice(timeout)
 }
 
+// GetOperator returns the operator if avail, panic if the operator can't be found
```
favadi: it should be renamed to MustGetOperator.

common/blockchain/base_blockchain.go
favadi: If we named the function `self.MustGetOperator` we can remove the == nil handling. Otherwise there is no guarantee operator will be != nil looking at the codes.

common/types.go
favadi: It is possible to create a invalid Timestamp instance, we can't have control of everything, think of reserve-data/common as a package other packages can import. This function should be named `ToUint64` or something. [gorename](https://godoc.org/golang.org/x/tools/cmd/gorename) might be helpful here.

exchange/bittrex.go
```diff
@@ -226,7 +226,7 @@ func (self *Bittrex) Withdraw(token common.Token, amount *big.Int, address ether
 	return "", errors.New(resp.Error)
 }
 
-func bitttimestampToUint64(input string) uint64 {
+func bitttimestampToUint64(input string) (uint64, error) {
```
favadi: this function name looks weird, maybe it should be renamed to `bitrexTimestampToUint64`?

exchange/bittrex/bittrex_endpoint.go
```diff
@@ -31,6 +31,7 @@ func nonce() string {
 
 func addPath(original string, path string) string {
 	url, err := url.Parse(original)
+	// If parsing raises error, this mean configuration is wrong.
```
favadi: this function should be named mustAddPath?

exchange/stable_ex.go
favadi: If user called a HTTP API and the server panic, it is wrong. If that is the case here, I'm on @cyberliem side.

settings/storage/setting_bolt.go
favadi: Please make sure the caller panic in this case. Please paste here the exact line the caller of this function handle the error.

stat/storage/analytic_bolt.go
favadi: Please paste here the exact line the caller of this function handle the error.

stat/storage/analytic_bolt.go
```diff
@@ -40,7 +40,7 @@ func NewBoltAnalyticStorage(dbPath string) (*BoltAnalyticStorage, error) {
 		return nil
 	})
 	if err != nil {
-		panic(err)
+		return nil, err
```
favadi: Please paste here the exact line the caller of this function handle the error.

stat/storage/fee_setrate_bolt.go
```diff
@@ -33,7 +33,7 @@ func NewBoltFeeSetRateStorage(path string) (*BoltFeeSetRateStorage, error) {
 	var db *bolt.DB
 	db, err = bolt.Open(path, 0600, nil)
 	if err != nil {
-		panic(err)
+		return nil, err
```
favadi: Please paste here the exact line the caller of this function handle the error.

stat/util/ip_locator.go
```diff
@@ -17,7 +17,8 @@ func IPToCountry(ip string) (string, error) {
 	il, err := newIPLocator(dbPath)
 	if err != nil {
 		// MaxMind's DB is stored in source tree
-		panic(err)
+		return "", fmt.Errorf("cannot get IPlocator %s", err.Error())
```
favadi: this should be panic because it can only happen if the developers remove, change the maxmind db in source tree to something invalid, so ProgrammerError.

stat/storage/analytic_bolt.go
cyberliem: https://github.com/KyberNetwork/reserve-data/blob/7f97ed3dc836906a8cb0e335d37dfc8cc42f77c7/cmd/configuration/getconfig.go#L108
Here, any error occur in creating setting DB will panic.

stat/storage/analytic_bolt.go
cyberliem: https://github.com/KyberNetwork/reserve-data/blob/7f97ed3dc836906a8cb0e335d37dfc8cc42f77c7/cmd/configuration/config.go#L150
at stat object init, any db creation failed will panic .

stat/storage/analytic_bolt.go
cyberliem: https://github.com/KyberNetwork/reserve-data/blob/7f97ed3dc836906a8cb0e335d37dfc8cc42f77c7/cmd/configuration/config.go#L150
at stat object init, any db creation failed will panic .

stat/storage/fee_setrate_bolt.go
cyberliem: https://github.com/KyberNetwork/reserve-data/blob/7f97ed3dc836906a8cb0e335d37dfc8cc42f77c7/cmd/configuration/config.go#L150
similar to those above.

PR #377, opened 2018-07-26
init IP locator from stat

cyberliem:


cyberliem:
[IP locator should be init from server start, instead of calling it in every IPtoCountry function.](https://trello.com/c/DtQ7gZVJ/192-ip-locator-should-be-init-from-server-start-instead-of-calling-it-in-every-iptocountry-function)

PR #388, opened 2018-07-30
changed blockchain object of http to interface

cyberliem:


cyberliem:
[change HTTP's blockchain to interface.](https://trello.com/c/fjVs0xJA/198-change-https-blockchain-to-interface)

http/http_blockchain.go
```diff
@@ -0,0 +1,10 @@
+package http
+
+import (
+	ethereum "github.com/ethereum/go-ethereum/common"
+)
+
+type HTTPBlockchain interface {
```
favadi: Please add comment doc and rename the interface to just `Blockchain`. It is considered bad practice to prefix the type with package name.

> http_blockchain.go:7:6: type name will be used as http.HTTPBlockchain by other packages, and that stutters; consider calling this Blockchain

PR #390, opened 2018-07-30
fixed nil kyber exchanges

cyberliem:


cyberliem:
[Core failed to start when KYBER_EXCHANGES=""](https://trello.com/c/5w4oJfYG/199-core-failed-to-start-when-kyberexchanges)

settings/exchange_setting.go
```diff
@@ -35,12 +35,19 @@ var exchangeNameValue = map[string]ExchangeName{
...
+// Running Exchange get the exchangeEnvironment params and return the list of exchanges ID for the current run
+// It returns empty string slice if the ENV is empty string or not found
+// DO NOT CALL this once httpserver has ran.
 func RunningExchanges() []string {
 	exchangesStr, ok := os.LookupEnv(exchangeEnv)
 	if !ok {
 		log.Print("WARNING: core is running without exchange")
 		return nil
 	}
+	if len(exchangesStr) == 0 {
```
favadi: the log when !ok and len(exchangesStr) == 0 is nearly identical. Is there any reason we shouldn't handle both of them in same manner?

cyberliem: It's totally different. Consider the case: 
`./cmd server` 
and the case : 
`KYBER_EXCHANGE="" ./cmd server`

The first case meaning the required param KYBER_EXCHANGE is not available. The second case mean  the KYBER_EXCHANGE is purposely set to be empty.

PR #393, opened 2018-07-31
only return Intermediator if huobi is avail

cyberliem:


cyberliem:
[/setting/all-settings returns 500 if huobi exchange is disabled](https://trello.com/c/Xhz3GSqn/203-setting-all-settings-returns-500-if-huobi-exchange-is-disabled)

PR #394, opened 2018-07-31
fix nil exchangefee panic when update exchange fee

cyberliem:


cyberliem:
[Fix nil exchange fee panic when update exchange fee.](https://trello.com/c/pv25IpNP/205-fix-nil-exchange-fee-panic-when-update-exchange-fee)

PR #395, opened 2018-07-31
allow token update when exchange setting is not in DB

cyberliem:


cyberliem:
[Update a token with Exchange is allowed when currently the exchange is empty](https://trello.com/c/wwunR1cc/207-update-a-token-with-exchange-is-allowed-when-currently-the-exchange-is-empty)

PR #396, opened 2018-07-31
Test setting apis

cyberliem:


common/test_exchange_for_setting.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"errors"
+	"math/big"
+
+	ethereum "github.com/ethereum/go-ethereum/common"
+)
+
+const BinanceTestExchangeID = "binance"
+
+func AddTestExchangeForSetting() {
```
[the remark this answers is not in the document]
favadi: Also, it might cause a race-condition bug when this function is called in tests that run in parallel.

http/exchange_update_test.go
```diff
@@ -0,0 +1,329 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	setting, err := settings.NewSetting(tokenSetting, addressSetting, exchangeSetting)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	storage, err := storage.NewBoltStorage(filepath.Join(tmpDir, "test.db"))
```
favadi: please  rename `storage` as it conflicts with package name imported above.

http/token_update_test.go
```diff
@@ -0,0 +1,470 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	setting, err := settings.NewSetting(tokenSetting, addressSetting, exchangeSetting)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	storage, err := storage.NewBoltStorage(filepath.Join(tmpDir, "test.db"))
```
favadi: please rename storage variable.

http/token_update_test.go
```diff
@@ -0,0 +1,470 @@
...
+			assert: httputil.ExpectSuccess,
+		},
+	}
+	for _, tc := range tests {
+		t.Run(tc.msg, func(t *testing.T) { testHTTPRequest(t, tc, testServer.r) })
+	}
+
+}
+
+type TestHTTPBlockchain struct {
```
favadi: could this be unexported?

settings/storage/exchange_bolt.go
```diff
@@ -21,7 +21,7 @@ func (boltSettingStorage *BoltSettingStorage) GetFee(ex settings.ExchangeName) (
 		}
 		data := b.Get(boltutil.Uint64ToBytes(uint64(ex)))
 		if data == nil {
-			return fmt.Errorf("key %s hasn't existed yet", ex.String())
+			return settings.ErrExchangeRecordNotFound
```
favadi: as the error now doesn't contain the key anymore, mabe we should add a log.Printf here with it?

favadi: same for everywhere after this.

http/settings_handler.go
```diff
@@ -51,19 +51,36 @@ func (self *HTTPServer) ensureRunningExchange(ex string) (settings.ExchangeName,
 func (self *HTTPServer) getExchangeSetting(exName settings.ExchangeName) (*common.ExchangeSetting, error) {
 	exFee, err := self.setting.GetFee(exName)
 	if err != nil {
-		return nil, err
+		if err != settings.ErrExchangeRecordNotFound {
+			return nil, err
+		}
+		log.Printf("the current exchange fee for %s hasn't existed yet. Overwrite new setting instead", exName.String())
```
favadi: does it make sense that the function name is `getExchangeSetting` but we are talking about overwrite here?

favadi:
[Add setting API tests.](https://trello.com/c/sI8Zrpr0/195-add-setting-api-tests)

common/test_exchange_for_setting.go
```diff
@@ -0,0 +1,85 @@
+package common
```
favadi: the file is still here?

common/testutil/test_exchange_for_setting.go
```diff
@@ -0,0 +1,89 @@
...
+	"errors"
+	"math/big"
+
+	"github.com/KyberNetwork/reserve-data/common"
+	ethereum "github.com/ethereum/go-ethereum/common"
+)
+
+const BinanceTestExchangeID = "binance"
+
+func EnsureTestUtilInit() bool {
```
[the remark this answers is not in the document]
cyberliem: As explained yesterday, since the test doesn't use AddSupportedExchange anymore, there must be a function for token_update_test to call to be able to import testutil [here](https://github.com/KyberNetwork/reserve-data/pull/396/files#diff-275173c1b4887831d5e7acc85abcceadR354) . If this function call isn't here, golang will remove testutil from imported packages of token_update_test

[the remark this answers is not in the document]
cyberliem: After testing with unamed import alias, I found this function is no longer necessary.

favadi:
I pushed 9dfa7019cf46899867e0d3c02eb21136c35efc13.

cyberliem:
[There are two TestExchange: one in common and one in reserve-core. This can be duplicated](https://trello.com/c/HqwvAB6J/153-there-are-two-testexchange-one-in-common-and-one-in-reserve-core-this-can-be-duplicated)

PR #402, opened 2018-08-01
use setting in stat init

cyberliem:


cyberliem:
[reserve Stat got a nil setting interface](https://trello.com/c/6Cs515Bs/217-reserve-stat-got-a-nil-setting-interface)

cyberliem:
[Stat is reading token setting directly from database instead of using core API](https://trello.com/c/Z9ARAuZB/210-stat-is-reading-token-setting-directly-from-database-instead-of-using-core-api)

PR #403, opened 2018-08-02
Check auth data after token update

cyberliem:


cyberliem:
[When confirm new settings. Core return in consistent authdata.](https://trello.com/c/1nBdxiAN/219-when-confirm-new-settings-core-return-in-consistent-authdata)

PR #404, opened 2018-08-02
Get cap by address

halink0803:


halink0803:
[Stat | implement "/daily-limit-by-address/:addr" API](https://trello.com/c/OMSuTpYh/214-stat-implement-daily-limit-by-address-addr-api)

http/permission.go
```diff
@@ -3,8 +3,9 @@ package http
...
 const (
-	ReadOnlyPermission    Permission = iota // can only read data
-	RebalancePermission                     // can do everything except configure setting
-	ConfigurePermission                     // can read data and configure setting, cannot set rates, deposit, withdraw, trade, cancel activities
-	ConfirmConfPermission                   // can read data and confirm configuration proposal
+	ReadOnlyPermission      Permission = iota // can only read data
+	RebalancePermission                       // can do everything except configure setting
+	ConfigurePermission                       // can read data and configure setting, cannot set rates, deposit, withdraw, trade, cancel activities
+	ConfirmConfPermission                     // can read data and confirm configuration proposal
+	UserDashboardPermission                   // can push kyc info to reserve stat
```
favadi: Please reuse the ConfirmConfPermission permission for the new KYC API, as the consumer is already used it for the `/update-user-addresses` API.

stat/reserve_stat.go
```diff
@@ -390,18 +395,43 @@ func (self ReserveStats) Stop() error {
...
+//GetTxCapByAddress return user Tx limit by wei
+func (rs ReserveStats) GetTxCapByAddress(addr ethereum.Address) (*big.Int, error) {
+	address := common.AddrToString(addr)
+	kyced, err := rs.GetKYCByAddress(address)
+	if err != nil {
+		return nil, err
+	}
+	var cap float64
+	if kyced {
+		cap = common.KycedCap().TxLimit
```
favadi: cap is a builtin keyword, could you please change the variable to something else?

http/permission.go
halink0803: @favadi I removed in next commit. However, as I checked again with user dashboard, they only call this api for core and have not use any key to call any api for core before. So, please confirm if this key is ConfirmConfPermission.

halink0803:
There is another PR https://github.com/KyberNetwork/reserve-data/pull/412 was merged into master so I will close this PR.

PR #405, opened 2018-08-02
versioned all setting

cyberliem:


cyberliem:
[When confirm new settings, there is a need to sync between new setting and auth data](https://trello.com/c/1nBdxiAN/219-when-confirm-new-settings-there-is-a-need-to-sync-between-new-setting-and-auth-data)

PR #407, opened 2018-08-06
Enable race condition detector for test in travis

halink0803:


halink0803:
[Enable race condition detector for test](https://trello.com/c/LwnG9JZU/132-enable-race-condition-detector-for-test)

PR #411, opened 2018-08-10
export and prune data according to agreed format

cyberliem:


cyberliem:
[Control rates.db from stat](https://trello.com/c/TDI6tg9F/222-control-ratesdb-from-stat)

stat/stat_storage_control.go
```diff
@@ -0,0 +1,154 @@
...
+				}
+			}
+		}
+		if err := os.Remove(fileName); err != nil {
+			log.Fatal(err)
+		}
+	}
+}
+
+// uploadAndVerify upload the file to remote storage and check its intergrity
```
halink0803: `integrity` not `intergrity`

stat/stat_storage_control.go
```diff
@@ -0,0 +1,154 @@
...
+		t := <-self.storageController.Runner.GetRateStorageControlTicker()
+		timepoint := common.TimeToTimepoint(t)
+		log.Printf("StatPruner: got signal in rate storage control channel with timestamp %d", timepoint)
+		//continuously pruning until there is no more expired data.
+		for {
+			tempfileName := fmt.Sprintf("./exported/TempExpireRateData.gz")
+			fromTime, toTime, nRecord, err := self.rateStorage.ExportExpiredRateData(common.GetTimepoint(), tempfileName)
+			fileName := fmt.Sprintf("./exported/ExpiredRateData_%d_%d.gz", fromTime, toTime)
+			if rErr := os.Rename(tempfileName, fileName); rErr != nil {
+				log.Printf("StatPruner: cannot rename file (%s)", rErr)
```
halink0803: Is this error acceptable? I see you print error and stop the loop, but not return the error.

stat/stat_storage_control.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func (self ReserveStats) RunStorageController() error {
+	err := self.storageController.Runner.Start()
+	if err != nil {
+		return err
+	}
+	go func() {
+		if cErr := self.ControllPriceAnalyticSize(); cErr != nil {
+			log.Printf("Control price analytic failed: %s", cErr.Error())
```
halink0803: Should return cErr?

stat/stat_storage_control.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}()
+	go func() {
+		if rErr := self.ControllRateSize(); rErr != nil {
+			log.Printf("StatPruner: Control rate analytic failed: %s", rErr.Error())
```
halink0803: Should it returns rErr?

stat/statpruner/storage_controller_runner.go
```diff
@@ -7,14 +7,17 @@ import (
...
+	GetRateStorageControlTicker() <-chan time.Time
 	Start() error
 	Stop() error
 }
 
 type ControllerTickerRunner struct {
-	ascduration time.Duration
-	ascclock    *time.Ticker
-	signal      chan bool
+	ascduration  time.Duration
```
halink0803: should be ascDuration.

stat/statpruner/storage_controller_runner.go
```diff
@@ -7,14 +7,17 @@ import (
...
+	rateDuration time.Duration
+	ascclock     *time.Ticker
```
halink0803: should it be ascClock?

stat/storage/rate_bolt.go
```diff
@@ -88,3 +100,101 @@ func (self *BoltRateStorage) GetReserveRates(fromTime, toTime uint64, ethReserve
...
+		if cErr := zw.Close(); cErr != nil {
+			log.Printf("StatPruner: closing gzip error %s", cErr.Error())
+		}
+		if cErr := outFile.Close(); cErr != nil {
+			log.Printf("Expire file close error: %s", cErr.Error())
+		}
+	}()
+
+	err = self.db.View(func(tx *bolt.Tx) error {
+		var uErr error
```
halink0803: inside db View function, should use only vErr for consistent. Below also have bUerr which is not very clear.

halink0803:
From now on, new function should not use `self` or `this`.

stat/stat_storage_control.go
cyberliem: prune storage is a continuous process, if the program return now it will break that logic.
In practical, we has a lot of error returned from the pruning process (network down etc). The record from S3 shows that there was error, and the prune continue the next day, success and prune records for 2 days in the past.

stat/stat_storage_control.go
cyberliem: same as above

halink0803:
PR https://github.com/KyberNetwork/reserve-data/pull/413 was merged into master, so this PR could be close.

cyberliem:
This pr is duplicated from [#413](https://github.com/KyberNetwork/reserve-data/pull/413)

PR #412, opened 2018-08-10
Update api to get user cap for each TX by address

halink0803:


halink0803:
[Stat | implement "/daily-limit-by-address/:addr" API](https://trello.com/c/OMSuTpYh/214-stat-implement-daily-limit-by-address-addr-api)

PR #414, opened 2018-08-13
Update api /cap-by-address

halink0803:


halink0803:
[Change user kyced tx limit](https://trello.com/c/APmRnFND/228-change-user-kyced-tx-limit)

PR #420, opened 2018-08-22
Add get step function data

halink0803:
TODO tasks:

- [ ] ~Wrap Blockchain ContractCaller to match abi/bind.ContractCaller interface~
- [x] Concurrent getStepFunction
- [x] Prefetch step function data

I change to use our Contract.Call function, so don't need to wrap our blockchain contract caller anymore.

### DIFF
--- a/data/fetcher/fetcher.go
+++ b/data/fetcher/fetcher.go
@@ -752,23 +752,31 @@ func (fc *Fetcher) RunStepFunctionDataStorage() {
 	tick := time.NewTicker(1 * time.Minute)
 	for {
 		data := sync.Map{}
+		// get current blockchain block
+		block, err := fc.blockchain.CurrentBlock()
+		if err != nil {
+			log.Printf("Cannot get current block: %s", err.Error())
+		}
+
 		// fetch all data from blcokchain
-		err := fc.fetchStepFunctionData(&data)
+		err = fc.fetchStepFunctionData(&data, block)
 
 		if err != nil {
 			log.Printf("fetch step function data error: %s", err.Error())
 		} else {
 			// convert data from sync map to go object
 			result := common.StepFunctionData{}
+			result.BlockNumber = block
+			result.Tokens = map[string]common.StepFunctionResponse{}
+
 			data.Range(func(k, v interface{}) bool {
 				token := k.(string)
 				stepData := v.(common.StepFunctionResponse)
-				result[token] = stepData
+				result.Tokens[token] = stepData
 				return true
 			})
 
 			// save data to storage
-			log.Printf("Result length: %d", len(result))
 			err := fc.stepFunctionDataStorage.StoreStepFunctionData(result)
 			if err != nil {
 				log.Printf("Store step function data error: %s", err.Error())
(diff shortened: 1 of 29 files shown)


halink0803:
TODO:
- [ ] Check valid data from blockchain

halink0803:
[Create an endpoint which return all imbalanceStepFunc of supported token](https://trello.com/c/TGXqBomo/234-create-an-endpoint-which-return-all-imbalancestepfunc-of-supported-token)

data/fetcher/fetcher.go
```diff
@@ -743,3 +747,87 @@ func (self *Fetcher) fetchPriceFromExchange(wg *sync.WaitGroup, exchange Exchang
...
+			data.Range(func(k, v interface{}) bool {
+				token := k.(string)
+				stepData := v.(common.StepFunctionResponse)
+				result.Tokens[token] = stepData
+				return true
+			})
+
+			// save data to storage
+			err := fc.stepFunctionDataStorage.StoreStepFunctionData(result)
+			if err != nil {
```
favadi: using same error instance as outer scope.

data/fetcher/fetcher.go
```diff
@@ -743,3 +747,87 @@ func (self *Fetcher) fetchPriceFromExchange(wg *sync.WaitGroup, exchange Exchang
...
+	}
+}
+
+func (fc *Fetcher) fetchStepFunctionData(data *sync.Map, block uint64) error {
+	var fetchStepFunctionDataErrCh = make(chan error)
+
+	// get internal token list
+	tokens, err := fc.setting.GetInternalTokens()
+	if err != nil {
+		log.Printf("Cannot get internal tokens: %s", err.Error())
```
favadi: this should be a fatal error.

data/storage/bolt.go
```diff
@@ -1670,3 +1674,41 @@ func (self *BoltStorage) ConfirmTokenUpdateInfo(tarQty common.TokenTargetQtyV2,
...
+	err := bs.db.Update(func(tx *bolt.Tx) error {
+		dataJSON, uErr := json.Marshal(data)
+		if uErr != nil {
+			return uErr
+		}
+		b := tx.Bucket([]byte(stepFunctionBucket))
+		v := b.Get(key)
+		if v != nil {
+			uErr := b.Delete(key)
+			if uErr != nil {
```
favadi: I don't think it is necessary to delete the data before put the new key.

data/storage/bolt.go
```diff
@@ -1670,3 +1674,41 @@ func (self *BoltStorage) ConfirmTokenUpdateInfo(tarQty common.TokenTargetQtyV2,
...
+			}
+		}
+		return b.Put(key, dataJSON)
+	})
+	return err
+}
+
+//GetStepFunctionData return current step function data
+func (bs *BoltStorage) GetStepFunctionData() (common.StepFunctionData, error) {
+	key := []byte("latest_data")
```
favadi: this key should be a package level const.

favadi:
I pushed some minor fixes in https://github.com/KyberNetwork/reserve-data/pull/420/commits/3c2e67b52285d5d3ffb66d0c32bd82a67174a778.

PR #422, opened 2018-09-05
[WIP] switch to go modules

favadi:


favadi:
[Use another package management instead of Godep](https://trello.com/c/7st4TUzp/55-use-another-package-management-instead-of-godep)

favadi:
Close as deprecated.

PR #423, opened 2018-09-12
new gas price strategy

tranvictor:


favadi:
[Improve gas price setting scheme](https://trello.com/c/AIju9Xns/244-improve-gas-price-setting-scheme)

tranvictor:
@favadi just fixed null tx bug from your commit :D

PR #424, opened 2018-09-17
wait group done when encountered err at reserveSupportedToken

cyberliem:


cyberliem:
[/get-reserve-rate returns no data](https://trello.com/c/kl9elg1G/236-get-reserve-rate-returns-no-data)

stat/fetcher.go
```diff
@@ -751,6 +751,7 @@ func (self *Fetcher) FetchReserveRates(timepoint uint64) {
 			go self.GetReserveRates(block, reserve, tokens, &data, &wg)
 		} else {
 			log.Printf("ERROR: Can not get reserve rates %s", err)
```
favadi: minor suggestion, we should only add to the waitgroup right before the go routine function. 

```go
		tokens, err := self.ReserveSupportedTokens(reserve)
		if err == nil {
			wg.Add(1)
			go self.GetReserveRates(block, reserve, tokens, &data, &wg)
		} else {
			log.Printf("ERROR: Can not get reserve rates %s", err)
		}
```

PR #427, opened 2018-09-21
Add more error log

halink0803:
- Add more error log
- If country is empty, then set it to "unknown"

### DIFF
--- a/stat/fetcher.go
+++ b/stat/fetcher.go
@@ -41,6 +41,7 @@ const (
 
 	etherScanAPIEndpoint      = "http://api.etherscan.io/api"
 	broadcastKyberAPIEndpoint = "https://broadcast.kyber.network"
+	unknownCountry            = "unknown"
 )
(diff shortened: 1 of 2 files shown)


halink0803:
[/get-heat-map prod returns null](https://trello.com/c/qCHUf03u/248-get-heat-map-prod-returns-null)

stat/fetcher.go
```diff
@@ -857,11 +858,11 @@ func GetTradeGeo(ipLocator *statutil.IPLocator, txHash string) (string, string,
 		}
 		country, err = ipLocator.IPToCountry(response.Data.IP)
 		if err != nil {
-			return "", "", err
+			return "", unknownCountry, err
```
favadi: it is a bad practice to rely on a returning error when err != nil. I created a PR in https://github.com/KyberNetwork/reserve-data/pull/428/files.

PR #430, opened 2018-09-21
change exported to tmp for pruning

cyberliem:


cyberliem:
[exported folder should be put in .gitignore](https://trello.com/c/O6aOztHg/229-exported-folder-should-be-put-in-gitignore)

PR #440, opened 2018-09-24
verbose logging for set rate mined nonce

favadi:


favadi:
[Verbose logging for transaction nonce](https://trello.com/c/jrnE0vWx/252-verbose-logging-for-transaction-nonce)

PR #462, opened 2018-12-17
change logic of setrateminednonce

cyberliem:


common/blockchain/base_blockchain.go
```diff
@@ -76,6 +76,46 @@ func (self *BaseBlockchain) MustGetOperator(name string) *Operator {
...
+		nonceResults[i]++
+		if nonceResults[i] > maxPopularNonceCount {
+			maxPopularNonceCount = nonceResults[i]
+			maxPopularNonce = i
+			sameMaxcount = false
+
+		} else if nonceResults[i] == maxPopularNonceCount {
+			sameMaxcount = true
+		}
+	}
```
favadi: please add a log message with the nonceResults print out.

common/blockchain/base_blockchain.go
```diff
@@ -76,6 +76,46 @@ func (self *BaseBlockchain) MustGetOperator(name string) *Operator {
...
+				nonceChannel <- nonce.Uint64()
+			}
+		}(endpoint, client)
+	}
+	for i := range nonceChannel {
+		nonceResults[i]++
+		if nonceResults[i] > maxPopularNonceCount {
+			maxPopularNonceCount = nonceResults[i]
+			maxPopularNonce = i
+			sameMaxcount = false
```
favadi: while the logic looks correct, it could be confusing to read. Please refactor this to a function that a accept a `map[uint64][uint64]` and returns the most popular nonce and error is any. Write tests for it with at least following case.

- input: {1: 3, 2: 1, 3: 1} --> 3, nil
- input: {1: 2; 2: 2, 3: 1} --> 0, error about can't determine most popular mine.
- input {} --> error about no result

common/blockchain/base_blockchain.go
```diff
@@ -76,6 +76,46 @@ func (self *BaseBlockchain) MustGetOperator(name string) *Operator {
 	return op
 }
 
+//GetDominantMinedNonceFromAllNodes  return the nonce that is dominant from all nodes
+func (self *BaseBlockchain) GetDominantMinedNonceFromAllNodes(operator string) (uint64, error) {
+	var (
+		op                   = self.MustGetOperator(operator)
+		maxPopularNonce      uint64
+		maxPopularNonceCount uint64
```
favadi: mostPopular sounds better :).

common/blockchain/base_blockchain.go
```diff
@@ -76,6 +76,46 @@ func (self *BaseBlockchain) MustGetOperator(name string) *Operator {
 	return op
 }
 
+//GetDominantMinedNonceFromAllNodes  return the nonce that is dominant from all nodes
+func (self *BaseBlockchain) GetDominantMinedNonceFromAllNodes(operator string) (uint64, error) {
```
favadi: we need a test for this function, but with `t.Skip()` at the start of the test so we can manually turns it on but keep it off in CI.

PR #470, opened 2018-12-21
2011 add pending status and log for debugging

cyberliem:


data/fetcher/fetcher.go
```diff
@@ -352,63 +352,58 @@ func (self *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord)
 			}
 			status, blockNum, err = self.blockchain.TxStatus(tx)
 			if err != nil {
-				log.Printf("Getting tx status failed, tx will be considered as pending: %s", err)
-			}
-			switch status {
-			case "":
-				if nonceValidator(activity) {
+				log.Printf("TX_STATUS: Getting tx status failed, tx will be considered as pending: %s", err)
```
favadi: from I can see, change the function signature to returns an error is better, as the consumer codes https://github.com/KyberNetwork/reserve-data/blob/4ebb8e449e18d5c82b9b00ab8055fa33297f8cb6/data/fetcher/fetcher.go#L265 does has the ability to deal with the returning error.

data/fetcher/fetcher.go
```diff
@@ -343,77 +352,71 @@ func (self *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord)
...
 			}
 			status, blockNum, err = self.blockchain.TxStatus(tx)
 			if err != nil {
-				log.Printf("Getting tx status failed, tx will be considered as pending: %s", err)
-			}
-			switch status {
-			case "":
-				if nonceValidator(activity) {
+				return result, fmt.Errorf("TX_STATUS: Getting tx status failed, tx will be considered as pending: %s", err)
+			} else {
```
favadi: we should remove the else clause as the err != nil case already returning the function.

data/fetcher/fetcher.go
```diff
@@ -343,31 +352,23 @@ func (self *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord)
 			var err error
 			txStr, ok := activity.Result["tx"].(string)
 			if !ok {
-				log.Printf("WARNING: cannot convert activity.Result[tx] (value %v) to string type", activity.Result["tx"])
-				continue
+				return result, fmt.Errorf("TX_STATUS: cannot convert activity.Result[tx] (value %v) to string type", activity.Result["tx"])
```
favadi: as I can see there are other types of pending activities feed to this function, it does not look safe to just abort execution here. Could you double check?

data/fetcher/fetcher.go
```diff
@@ -343,31 +352,23 @@ func (self *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord)
...
 			}
 			tx := ethereum.HexToHash(txStr)
 			if tx.Big().IsInt64() && tx.Big().Int64() == 0 {
 				continue
 			}
 			status, blockNum, err = self.blockchain.TxStatus(tx)
 			if err != nil {
-				log.Printf("Getting tx status failed, tx will be considered as pending: %s", err)
+				return result, fmt.Errorf("TX_STATUS: Getting tx status failed, tx will be considered as pending: %s", err)
```
favadi: do we really considered it as pending now? :)

data/fetcher/fetcher.go
cyberliem: > We should do the same as in `FetchBalanceFromBlockchain` function, returns an error, log that it failed to fetch and does nothing.

I think we should honour this logic, One activity crashed into Error, we'll return the whole execution.

data/fetcher/fetcher.go
cyberliem: fixed it

data/fetcher/fetcher.go
favadi: I'll give a better example.

- pendings contains 2 common.ActivityRecord
  - the first one has `activity.Result["tx"]` is an integer
  - the second one has `activity.Result["tx"]` is a string: https://github.com/KyberNetwork/reserve-data/blob/c24250af29cb4c6afa15774e546a0a0766e6fcaf/data/fetcher/fetcher.go#L344


With old logic:

- FetchStatusFromBlockchain ignores the first one and process the second one
- The first pending ActivityRecord can be processed by other function if needed

With this PR:

- FetchStatusFromBlockchain will terminate execution and returns an error until the first pending ActivityRecord is removed
- FetchStatusFromBlockchain

data/fetcher/fetcher.go
```diff
@@ -343,7 +351,7 @@ func (self *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord)
 			var err error
 			txStr, ok := activity.Result["tx"].(string)
 			if !ok {
-				log.Printf("WARNING: cannot convert activity.Result[tx] (value %v) to string type", activity.Result["tx"])
+				log.Printf("TX_STATUS: ERROR cannot convert activity.Result[tx] (value %v) to string type", activity.Result["tx"])
```
favadi: should we really need changing from WARNING to ERROR?

data/fetcher/fetcher.go
```diff
@@ -352,22 +360,16 @@ func (self *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord)
 			}
 			status, blockNum, err = self.blockchain.TxStatus(tx)
 			if err != nil {
-				log.Printf("Getting tx status failed, tx will be considered as pending: %s", err)
+				log.Printf("TX_STATUS: ERROR Getting tx status failed: %s", err)
```
favadi: why we don't return the error here?

PR #473, opened 2018-12-27
2070 Remove old stats code

halink0803:


blockchain/blockchain.go
```diff
@@ -390,43 +378,6 @@ func (self *Blockchain) FetchRates(atBlock uint64, currentBlock uint64) (common.
...
-		rateEntry.BuySanityRate = common.BigToFloat(sanityRate[index*2+1], 18)
-		rateEntry.SellReserveRate = common.BigToFloat(reserveRate[index*2], 18)
-		rateEntry.SellSanityRate = common.BigToFloat(sanityRate[index*2], 18)
-		result[fmt.Sprintf("ETH-%s", token.ID)] = rateEntry
-	}
-	rates.Data = result
-
-	return rates, err
-}
-
```
cyberliem: GetPricingMethod() is a method used in stat, but nolonger used. Please remove it

cyberliem: the same goes for GeneratedGetReserveRates()

cmd/configuration/config.go
```diff
@@ -82,12 +78,6 @@ type Config struct {
 	ActivityStorage      core.ActivityStorage
 	DataStorage          data.Storage
 	DataGlobalStorage    data.GlobalStorage
-	StatStorage          stat.StatStorage
```
cyberliem: Please remove the StoragePath for stats storage in SettingPaths objects and its related config as well. (Line 40)

cmd/clicmd/startserver_functions.go
```diff
@@ -213,43 +166,3 @@ func CreateDataCore(config *configuration.Config, kyberENV string, bc *blockchai
...
-	rStat := stat.NewReserveStats(
-		config.AnalyticStorage,
-		config.StatStorage,
-		config.LogStorage,
-		config.RateStorage,
-		config.UserStorage,
-		config.FeeSetRateStorage,
-		config.StatControllerRunner,
-		statFetcher,
-		config.Archive,
```
cyberliem: Please remove stat related function in archive package as well.
(for example: common/archive/archive.go line 18)

blockchain/blockchain.go
cyberliem: the same for InterpretTimestamp()

cyberliem: GetEthRate and GetUSDRate is not used for core anymore. I think we should remove the related codes.

common/types.go
```diff
@@ -595,148 +594,6 @@ type KNLog interface {
 	Type() string
 }
 
-type SetCatLog struct {
```
cyberliem: the whole user file is not used anymore, some type (for example, KNLog) is not used anymore.

PR #474, opened 2019-01-11
changed address to v3

cyberliem:


favadi:
Close as we won't maintain the old stats code anymore.

PR #483, opened 2019-01-24
2274 Update huobi deposit endpoint

halink0803:


exchange/huobi/huobi_endpoint.go
```diff
@@ -360,7 +360,7 @@ func (self *HuobiEndpoint) GetDepositAddress(asset string) (exchange.HuobiDeposi
 	result := exchange.HuobiDepositAddress{}
 	respBody, err := self.GetResponse(
 		"GET",
-		self.interf.AuthenticatedEndpoint()+"/v1/dw/deposit-virtual/addresses",
+		"https://www.hbg.com/-/x/pro/"+"v1/dw/deposit-virtual/addresses",
```
favadi: I feel like we could just change the authenticated endpoint to "https://www.hbg.com/-/x/pro/", could you check?

halink0803: I think we should not set authenticated endpoint to "hbg.com/-/x/pro" because it is not officially mention in huobi docs. This api was inspected from huobi website. I think we can just keep it like current.

favadi: fair point, just make it a const somewhere and we can merge.

PR #495, opened 2019-04-08
1503 remove json env setting files

cyberliem:


cmd/configuration/address_config.go
```diff
@@ -0,0 +1,23 @@
+package configuration
+
+import (
+	"github.com/KyberNetwork/reserve-data/common"
+)
+
+//AddressConfigs store token configs according to env mode.
+var AddressConfigs = map[string]string{
+	common.DevMode: `
```
favadi: I don't think we should keep those values in a string. Please use a proper type for it.

cmd/configuration/exchange_config.go
```diff
@@ -0,0 +1,57 @@
+package configuration
+
+import (
+	"github.com/KyberNetwork/reserve-data/common"
+)
+
+//ExchangeConfigs store exchange config according to env mode.
+var ExchangeConfigs = map[string]string{
+	common.DevMode: `{
```
favadi: same comment as above.

PR #540, opened 2019-07-29
valid-update-asset-params

ducquangkstn:


v3/http/asset_update.go
```diff
@@ -27,6 +35,42 @@ func (s *Server) createUpdateAsset(c *gin.Context) {
 	httputil.ResponseSuccess(c, httputil.WithField("id", id))
 }
 
+func (s *Server) checkUpdateAssetParams(updateEntry common.UpdateAssetEntry) error {
+	access, err := s.storage.GetAsset(updateEntry.AssetID)
```
secmask: it's `asset`

v3/http/asset_update.go
```diff
@@ -27,6 +35,42 @@ func (s *Server) createUpdateAsset(c *gin.Context) {
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to get asset id: %v from db ", updateEntry.AssetID)
+	}
+
+	if updateEntry.Rebalance != nil {
```
secmask: we can combine these 2 `if`

v3/http/asset_update.go
```diff
@@ -27,6 +35,42 @@ func (s *Server) createUpdateAsset(c *gin.Context) {
...
+		if *updateEntry.SetRate != common.SetRateNotSet && access.PWI == nil && updateEntry.PWI == nil {
+			return common.ErrPWIMissing
+		}
+	}
+
+	if updateEntry.Transferable != nil {
+		if *updateEntry.Transferable {
+			for _, exchange := range access.Exchanges {
+				if common.IsZeroAddress(exchange.DepositAddress) {
+					return common.ErrDepositAddressMissing
```
secmask: it's useful if we can show detail on which Exchange the Deposit Address is missing. And because this a batch API, we should also detail which assetID we're working on.

v3/http/asset_update_test.go
```diff
@@ -321,3 +367,104 @@ func TestCreateUpdateAsset(t *testing.T) {
...
+
+	const updateAsset = "/v3/update-asset"
+	setRate1 := common.GoldFeed
+	var tests = []testCase{
+		{ // pwi in server is not nil
+			data: common.CreateUpdateAsset{
+				Assets: []common.UpdateAssetEntry{
+					{
+						AssetID: assetID,
+						SetRate: &setRate1,
```
secmask: tip, there's also common.SetRatePointer.

PR #546, opened 2019-07-31
4291 validate trading pair

ducquangkstn:


v3/http/trading_pair_create.go
```diff
@@ -27,6 +36,44 @@ func (s *Server) createCreateTradingPair(c *gin.Context) {
...
+	quote, err := s.storage.GetAsset(createEntry.Quote)
+	if err != nil {
+		return errors.Wrapf(common.ErrBaseAssetInvalid, "quote id: %v", createEntry.Quote)
+	}
+
+	if !quote.IsQuote {
+		return errors.Wrap(common.ErrQuoteAssetInvalid, "quote asset should have is_quote=true")
+	}
+
+	containExchange := false
```
secmask: this block can be extract to function

v3/http/trading_pair_create.go
```diff
@@ -27,6 +36,44 @@ func (s *Server) createCreateTradingPair(c *gin.Context) {
...
+	for _, exchange := range base.Exchanges {
+		if exchange.ExchangeID == createEntry.ExchangeID {
+			containExchange = true
+		}
+	}
+	if !containExchange {
+		return errors.Wrap(common.ErrBaseAssetInvalid, "exchange id not found")
+	}
+
+	containExchange = false
```
secmask: this finding block can also extract into function

v3/http/trading_pair_create.go
ducquangkstn: change as request

v3/http/trading_pair_create.go
ducquangkstn: change as request

PR #552, opened 2019-08-05
implement get trading live info when create asset

secmask:


v3/http/asset_create.go
```diff
@@ -1,15 +1,105 @@
...
+					tradingPairSymbol.BaseSymbol = assetExchange.Symbol
+					quote, err := getAssetExchange(assets, tp.Quote, assetExchange.ExchangeID)
+					if err != nil {
+						return err
+					}
+					tradingPairSymbol.QuoteSymbol = quote.Symbol
+				}
+				tps = append(tps, tradingPairSymbol)
+			}
+			for _, exchange := range s.exchanges {
```
ducquangkstn: pls use common.SupportedExchanges instead of

PR #559, opened 2019-08-06
update pending activities

halink0803:


halink0803:
@secmask please check

PR #570, opened 2019-08-08
Update binding tags

halink0803:
Because gin `required` tag does not accept zero value, however, in many apis we have `id` field required, which is started from 0. The solution is using binding tag `exists` instead of `required` (https://github.com/gin-gonic/gin/issues/491)

### DIFF
--- a/v3/http/trading_by_create.go
+++ b/v3/http/trading_by_create.go
@@ -56,7 +56,7 @@ func (s *Server) getCreateTradingBys(c *gin.Context) {
 
 func (s *Server) getCreateTradingBy(c *gin.Context) {
 	var input struct {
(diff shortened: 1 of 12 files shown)


v3/http/trading_by_create.go
```diff
@@ -56,7 +56,7 @@ func (s *Server) getCreateTradingBys(c *gin.Context) {
 
 func (s *Server) getCreateTradingBy(c *gin.Context) {
 	var input struct {
-		ID uint64 `uri:"id" binding:"required"`
+		ID uint64 `uri:"id" binding:"exists"`
```
secmask: why we use exists for `tradingBy`?

PR #573, opened 2019-08-12
Add listen flag, remove port flag

halink0803:
Current portFlag not allow core to bind to other host different from localhost which does not work with gateway. Change using listen flag.

### DIFF
--- a/cmd/configuration/flags.go
+++ b/cmd/configuration/flags.go
@@ -1,6 +1,7 @@
 package configuration
 
 import (
+	"fmt"
 	"log"
 
 	ethereum "github.com/ethereum/go-ethereum/common"
@@ -20,7 +21,7 @@ import (
 )
 
 const (
-	portFlag         = "port"
+	httpAddressFlag  = "listen"
 	portDefaultValue = 8000
 
 	dryRunFlag = "dry-run"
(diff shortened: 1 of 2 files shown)


cmd/configuration/flags.go
```diff
@@ -104,36 +105,36 @@ func NewDryRunFromContext(c *cli.Context) bool {
...
-	return cli.IntFlag{
-		Name:   portFlag,
-		Usage:  "server port",
-		EnvVar: "PORT",
-		Value:  portDefaultValue,
+//NewHTTPAddressFlag new flag for http address
+func NewHTTPAddressFlag() cli.Flag {
+	return cli.StringFlag{
+		Name:   httpAddressFlag,
+		Usage:  "http address",
```
secmask: suggest to use `Usage` as `"bind address for http interface"`

cmd/configuration/flags.go
```diff
@@ -20,7 +21,7 @@ import (
 )
 
 const (
-	portFlag         = "port"
+	httpAddressFlag  = "listen"
```
secmask: suggest to rename `listen` -> `addr`,( in case we have multiple listen interface for different protocol, we can make them like `http-addr`, `grpc-addr` ), `listen` is a verb.

secmask:
@vietthang207 cc

PR #597, opened 2019-08-19
move price-factor to postgres

secmask:


v3/storage/postgres/sql.go
```diff
@@ -143,6 +143,12 @@ CREATE TABLE IF NOT EXISTS setting_change (
     data JSON NOT NULL
 );
 
+CREATE TABLE IF NOT EXISTS price_factor (
```
ducquangkstn: pls check whether we should create a field store all json data or split into multiple row and have reference to assetID

ducquangkstn: may be just create it as a todo job and fix it later

v3/storage/postgres/sql.go
```diff
@@ -143,6 +143,12 @@ CREATE TABLE IF NOT EXISTS setting_change (
...
+    id serial primary key,
+  	timepoint bigint NOT NULL,
```
ducquangkstn: it should be changed to timestamp type

v3/storage/interface.go
```diff
@@ -28,6 +28,9 @@ type Interface interface {
 	GetSettingChanges() ([]v3.SettingChangeResponse, error)
 	RejectSettingChange(uint64) error
 	ConfirmSettingChange(uint64, bool) error
+
+	CreatePriceFactor(v3.PriceFactorAtTime) (uint64, error)
+	GetPriceFactors(uint64, uint64) ([]v3.PriceFactorAtTime, error)
```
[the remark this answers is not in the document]
secmask: we can do it later

v3/storage/postgres/sql.go
secmask: Luke Tran said we only need to store and receive function for this, it actually does not relate to asset feature, just left it as is.

v3/storage/postgres/sql.go
secmask: it usually a time, but not limit to, this will be fill by caller, and not guarantee to be a valid timestamp, so I decide to convert it from timestamp to bigint.

PR #610, opened 2019-08-22
setting change catalog update exchange

ducquangkstn:


secmask:
LGTM, but we also need to update `gateway` to support these new handler

PR #629, opened 2019-08-26
Add check token indice in create asset params

halink0803:


v3/http/setting_change.go
```diff
@@ -440,6 +440,11 @@ func getAssetExchange(assets []common.Asset, assetID, exchangeID uint64) (common
 }
 
 func (s *Server) checkCreateAssetParams(createEntry common.CreateAssetEntry) error {
+	if createEntry.Transferable {
+		if err := s.blockchain.CheckTokenIndices(createEntry.Address); err != nil {
```
secmask: do we need to make sure `s.blockchain` is not nil?

v3/blockchain/flags.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	defaultRateAddress = map[deployment.Deployment]string{
+		deployment.Development: "0x798AbDA6Cc246D0EDbA912092A2a3dBd3d11191B",
+		deployment.Ropsten:     "0x535DE1F5a982c2a896da62790a42723A71c0c12B",
+		deployment.Staging:     "0xe3E415a7a6c287a95DC68a01ff036828073fD2e6",
+		deployment.Production:  "0x798AbDA6Cc246D0EDbA912092A2a3dBd3d11191B",
+	}
+	defaultNodeEndpoint = map[deployment.Deployment]string{
+		deployment.Development: "https://mainnet.infura.io/v3/fd330878eff84d48b97e3023c996dff6",
```
secmask: these values should be put in config because it contain credentials?

halink0803: added check for blockchain != nil.

v3/http/setting_change.go
halink0803: added check if blockchain == nil.

secmask:
@halink0803 please resolve conflict

halink0803:
> @halink0803 please resolve conflict

Conflict resolved.

reservesetting/blockchain/flags.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	defaultRateAddress = map[deployment.Deployment]string{
+		deployment.Development: "0x798AbDA6Cc246D0EDbA912092A2a3dBd3d11191B",
+		deployment.Ropsten:     "0x535DE1F5a982c2a896da62790a42723A71c0c12B",
+		deployment.Staging:     "0xe3E415a7a6c287a95DC68a01ff036828073fD2e6",
+		deployment.Production:  "0x798AbDA6Cc246D0EDbA912092A2a3dBd3d11191B",
+	}
+	defaultNodeEndpoint = map[deployment.Deployment]string{
+		deployment.Development: "https://mainnet.infura.io/",
```
secmask: we can make these value share somewhere and import both by reservesetting and fetcher for better consistent default value.

PR #638, opened 2019-08-28
support usd feed

secmask:


common/usd_data.go
```diff
@@ -0,0 +1,19 @@
+package common
+
+// USDData data return by /usd-feed
+type USDData struct {
+	Timestamp uint64
+	Coinbase  CoinbaseData `json:"CoincaseUSD"`
```
halink0803: CoinbaseUSD typo mistake.

PR #653, opened 2019-09-04
4572 add more test case

NNagato:


reservesetting/storage/postgres/exchanges_test.go
```diff
@@ -95,3 +95,19 @@ func TestStorage_UpdateExchange(t *testing.T) {
...
+func TestStorage_GetUpdateByName(t *testing.T) {
+	db, tearDown := testutil.MustNewDevelopmentDB()
+	defer func() {
+		assert.NoError(t, tearDown())
+	}()
+
+	s, err := NewStorage(db)
+	assert.NoError(t, err)
+	initData(t, s)
+	exchangeByID, err := s.GetExchange(1)
```
secmask: we have constant for `binance` exchange instead of `1`

NNagato: fixed in https://github.com/KyberNetwork/reserve-data/pull/653/commits/a2f38f4d0b8d7c102ad9e5d851a086b82b50baa7

PR #668, opened 2019-09-06
Hot fix set token update

ducquangkstn:


http/settings_handler.go
```diff
@@ -338,11 +341,11 @@ func (s *Server) RejectTokenUpdate(c *gin.Context) {
 // getInfosFromExchangeEndPoint assembles a map of exchange to lists of PairIDs and
 // query their exchange Info in one go
 func (s *Server) getInfosFromExchangeEndPoint(tokenUpdates map[string]common.TokenUpdate) (map[string]common.ExchangeInfo, error) {
-	const ETHID = "ETH"
 	exTokenPairIDs := make(map[string]([]common.TokenPairID))
 	result := make(map[string]common.ExchangeInfo)
 	for tokenID, tokenUpdate := range tokenUpdates {
-		if tokenUpdate.Token.Internal && tokenID != "ETH" {
+		// ETH-ETH pair is not exist
```
secmask: `does not exist` instead of `is not exist`

PR #679, opened 2019-09-10
add pprof

ducquangkstn:


cmd/clicmd/root.go
```diff
@@ -50,6 +50,7 @@ Allow overwriting some parameter`,
 	startServer.Flags().StringVar(&cliAddress.Pricing, "pricing-addr", defaultValue.Pricing, "pricing contract address")
 	startServer.Flags().StringVar(&cliAddress.Network, "network-addr", defaultValue.Network, "network contract address")
 	startServer.Flags().StringVar(&cliAddress.InternalNetwork, "internal-network-addr", defaultValue.InternalNetwork, "internal network contract address")
+	startServer.Flags().StringVar(&profilerPrefix, "profiler-prefix", "", "profiler prefix (disable profiler if empty)")
```
secmask: overall LGTM, but flag should give a hint how `profiler-prefix` look like, 
I would suggest description message to: set prefix for pprof http handler, eg "/debug/pprof", profiler will be disabled if this flag value is empty. A secure token can be put into profiler-prefix to limit access.

ducquangkstn: change as request

PR #681, opened 2019-09-10
Add cancel all order api

halink0803:
Always return success with form:

```json
{
    "success": true,
    "data": [
        {
            "id": 1,
            "reason": "binance key is invalid"
        }
    ]
}
```

if there is no failed cancel order then "data" will be empty.

### DIFF
--- a/http/exchange_api.go
+++ b/http/exchange_api.go
@@ -7,8 +7,16 @@ import (
 	"github.com/KyberNetwork/reserve-data/http/httputil"
 )
 
+type failedCancelOrder struct {
+	Reason string            `json:"reason"`
+	ID     common.ActivityID `json:"id"`
+}
+
 // CancelAllOrders cancel all orders
 func (s *Server) CancelAllOrders(c *gin.Context) {
+	var (
+		response []failedCancelOrder
+	)
 	pendingActivites, err := s.app.GetPendingActivities()
 	if err != nil {
 		httputil.ResponseFailure(c, httputil.WithError(err))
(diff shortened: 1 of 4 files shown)


http/exchange_api.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		httputil.ResponseFailure(c, httputil.WithError(err))
+		return
+	}
+
+	for _, activity := range pendingActivites {
+		if activity.Action == common.ActionTrade {
+			exchange := activity.Params["exchange"].(common.Exchange)
+			// Cancel order
+			if err := s.core.CancelOrder(activity.ID, exchange); err != nil {
```
secmask: if an activity status get changed during we process, this will abort all the next one, we can add a convenient param like 'on_error == continue' to ignore error?, and we can have result as a list of which success/fail. maybe we should ask Luke if he need this.

http/server.go
```diff
@@ -421,6 +421,7 @@ func (s *Server) register() {
 		g.GET("/immediate-pending-activities", s.ImmediatePendingActivities)
 
 		g.POST("/cancelorder", s.CancelOrder)
+		g.POST("/cancel-all-order", s.CancelAllOrders)
```
secmask: remember to add this to gateway.

http/exchange_api.go
halink0803: Yes, I think about this before. Let me ask him if there is an error, we stop or continue cancel other orders. Maybe we should cancel as many order as we can.

http/server.go
halink0803: Added.

http/exchange_api.go
halink0803: Updated

PR #695, opened 2019-09-17
fix loop not sleep as expected

ducquangkstn:


exchange/binance.go
```diff
@@ -295,8 +295,14 @@ func (bn *Binance) FetchOnePairTradeHistory(
 //FetchTradeHistory get all trade history for all tokens in the exchange
 func (bn *Binance) FetchTradeHistory() {
 	t := time.NewTicker(10 * time.Minute)
+	var isFirstLoop = true
 	go func() {
 		for {
+			if isFirstLoop { // if not first loop then sleep for 10 minutes
```
secmask: we can move this logic into `for` condition?

PR #699, opened 2019-09-18
Update assets list

halink0803:


secmask:
let's check other comment from Quang https://github.com/KyberNetwork/reserve-data/pull/698

halink0803:
> let's check other comment from Quang #698

I have changed follow his comment for appropriate ones.

PR #703, opened 2019-09-19
postgres to use timestamp with zone info

secmask:


secmask:
use `timestamp` without zone information can lead to wrong time data, eg, postgres will ignore timezone information when we insert data.

PR #709, opened 2019-09-25
collect body data in case feed error

secmask:


world/world.go
```diff
@@ -1,8 +1,11 @@
 package world
 
 import (
+	"bytes"
 	"encoding/json"
+	"errors"
```
ducquangkstn: because we haven't update to go 1.13 so pls change this to `https://github.com/pkg/errors` for a stack trace.

world/world.go
```diff
@@ -34,21 +37,23 @@ func (tw *TheWorld) getPublic(url string, dst interface{}) error {
...
+		errMsg := fmt.Sprintf("%s read response error %v", caller, err)
+		log.Println(errMsg)
+		return errors.New(errMsg)
+	}
 
 	if resp.StatusCode != http.StatusOK {
-		log.Printf("unexpected return code: %d", resp.StatusCode)
-		return fmt.Errorf("unexpected return code: %d", resp.StatusCode)
+		errMsg := fmt.Sprintf("unexpected return code: %d, response text %s", resp.StatusCode, body)
+		return errors.New(errMsg)
```
ducquangkstn: add `log.Println(errMsg)` to log the error

PR #722, opened 2019-10-16
add zap support

secmask:


data/fetcher/fetcher.go
```diff
@@ -784,19 +786,21 @@ func (f *Fetcher) FetchOrderbook(timepoint uint64) {
...
+		f.l.Warnf("Storing data failed: %", err)
 	}
 }
 
 func (f *Fetcher) fetchPriceFromExchange(wg *sync.WaitGroup, exchange Exchange, data *ConcurrentAllPriceData, timepoint uint64) {
 	defer wg.Done()
 	fetchConfig, err := f.globalStorage.GetAllFetcherConfiguration()
 	if err != nil {
-		log.Fatal("cannot get btc fetcher configuration")
+		f.l.Warnf("cannot get btc fetcher configuration err=%+v", err)
```
ducquangkstn: I recommend to keep the logic here, with f.l.Fatalf

secmask: it's safe to not fatal here, it's should be retry later, confirmed with Ha Link.

PR #726, opened 2019-10-18
clean up callcontract

secmask:


cmd/configuration/getconfig.go
```diff
@@ -137,23 +137,26 @@ func GetConfig(kyberENV string, authEnbl bool, endpointOW string, cliAddress com
...
+	var callClients []*common.EthClient
 	for _, ep := range bkEndpoints {
 		var bkClient *ethclient.Client
 		bkClient, err = ethclient.Dial(ep)
 		if err != nil {
 			l.Warnf("Cannot connect to %s, err %s. Ignore it.", ep, err)
 		} else {
 			bkClients[ep] = bkClient
-			callClients = append(callClients, bkClient)
+			callClients = append(callClients, &common.EthClient{
```
ducquangkstn: I recommend  to add a contructor method such as `NewEthClient(url string) (*EthClient, error)`. In addition, should we add a check if all backup client init failed

secmask: I updated.

PR #730, opened 2019-10-22
Master

secmask:


secmask:
this a back port patch from master.

PR #733, opened 2019-10-23
Set delisted token zero rate

halink0803:


http/server.go
```diff
@@ -253,6 +253,38 @@ func (s *Server) GetRate(c *gin.Context) {
...
+		// error might be from node, return just for warning
+		return nil, nil, nil, nil, err
+	}
+	if len(listedTokens) <= len(tokens) {
+		return tokens, bigBuys, bigSells, bigAfpMid, nil
+	}
+
+	for _, tokenAddr := range listedTokens {
+		exist := false
+		for _, token := range tokens {
```
secmask: we can extract this block to a function

PR #736, opened 2019-10-29
Update get huobi address api endpoint

halink0803:


exchange/huobi.go
```diff
@@ -48,7 +48,7 @@ func (h *Huobi) MarshalText() (text []byte, err error) {
 // It should only be used to send 2nd transaction.
 func (h *Huobi) RealDepositAddress(tokenID string) (ethereum.Address, error) {
 	liveAddress, err := h.interf.GetDepositAddress(tokenID)
-	if err != nil || liveAddress.Data[0].Address == "" {
+	if err != nil || len(liveAddress.Data) == 0 || liveAddress.Data[0].Address == "" {
 		h.l.Warnf("WARNING: Get Huobi live deposit address for token %s failed: (%v) or the replied address is empty. Check the currently available address instead", tokenID, err)
```
secmask: it's better to not let other guess  "A or B wrong"

PR #737, opened 2019-10-29
V3 - update huobi get deposit address endpoint

halink0803:


cmd/configuration/init_exchanges.go
```diff
@@ -127,7 +127,7 @@ func updateDepositAddress(assetStorage storage.Interface, be exchange.BinanceInt
 				err = assetStorage.UpdateDepositAddress(
 					asset.ID,
 					uint64(common.Huobi),
-					ethereum.HexToAddress(depositAddress.Address))
+					ethereum.HexToAddress(depositAddress.Data[0].Address))
```
secmask: need to check len `depositAddress.Data` before access.

PR #740, opened 2019-10-30
use err.Error() can lost detail in some case

secmask:


secmask:
zap/sentry can give more detail if we pass `err` instead of `err.Error()`

exchange/huobi.go
```diff
@@ -764,7 +764,7 @@ func (h *Huobi) WithdrawStatus(
 	}
 	withdraws, err := h.interf.WithdrawHistory(tokens)
 	if err != nil {
-		return "", "", fmt.Errorf("can't get withdraw history from huobi: %s", err.Error())
+		return "", "", fmt.Errorf("can't get withdraw history from huobi: %v", err)
```
ducquangkstn: change to errors.Wrap

http/settings_handler.go
```diff
@@ -407,7 +407,7 @@ func (s *Server) ensureInternalSetting(tokenUpdate common.TokenUpdate) error {
 	token := tokenUpdate.Token
 	if !token.IsETH() { // TokenIndices doesn't contains ETH
 		if uErr := s.blockchain.CheckTokenIndices(ethereum.HexToAddress(token.Address)); uErr != nil {
-			return fmt.Errorf("cannot get token indice from smart contract (%s) ", uErr.Error())
+			return fmt.Errorf("cannot get token indice from smart contract (%+v) ", uErr)
```
ducquangkstn: change to `errors.Wrap` pls

http/server.go
```diff
@@ -105,7 +105,7 @@ func eligible(ups, allowedPerms []Permission) bool {
 func (s *Server) Authenticated(c *gin.Context, requiredParams []string, perms []Permission) (url.Values, bool) {
 	err := c.Request.ParseForm()
 	if err != nil {
-		httputil.ResponseFailure(c, httputil.WithReason(fmt.Sprintf("Malformed request package: %s", err.Error())))
+		httputil.ResponseFailure(c, httputil.WithReason(fmt.Sprintf("Malformed request package: %v", err)))
```
ducquangkstn: According to what I have test, `%v` have same effect with `%s`, pls change it to `%+v`

http/settings_handler.go
secmask: ok

exchange/huobi.go
secmask: ok

http/server.go
secmask: %s and %v is the same if error is not point to a struct, otherwise it's different.

PR #747, opened 2019-11-12
Update get listed token once and update when listing new token

halink0803:


http/server.go
```diff
@@ -49,6 +49,7 @@ type Server struct {
 	blockchain     Blockchain
 	setting        Setting
 	l              *zap.SugaredLogger
+	listedTokens   []ethereum.Address
```
secmask: please add some comment on why we keep this list

http/settings_handler.go
```diff
@@ -410,6 +410,13 @@ func (s *Server) ensureInternalSetting(tokenUpdate common.TokenUpdate) error {
 		if uErr := s.blockchain.CheckTokenIndices(ethereum.HexToAddress(token.Address)); uErr != nil {
 			return pe.Wrap(uErr, "cannot get token indice from smart contract (%+v) ")
 		}
+
+		// update listed token for server
+		listedTokens, err := s.blockchain.GetListedTokens()
+		if err != nil {
+			return err
```
secmask: please add log because caller seem also did not to.

halink0803: added log: https://github.com/KyberNetwork/reserve-data/blob/3e98ac6df369bf55b5339331f0a7229f24cd1dad/http/settings_handler.go#L417

PR #749, opened 2019-11-13
Update binance endpoint to v3

halink0803:


exchange/huobi/huobi_endpoint.go
```diff
@@ -361,8 +361,6 @@ func (ep *Endpoint) GetDepositAddress(asset string) (exchange.HuobiDepositAddres
 			err = fmt.Errorf("get Huobi deposit address failed: %s", result.Message)
 		}
 	}
-	// log response for debugging
```
secmask: this should be put in a separate commit.

halink0803: Do you mean separate PR or just separate commit?

secmask: this is small change, so a separate commit is acceptable.

PR #758, opened 2019-11-19
update load and set tokens indices

halink0803:


cmd/clicmd/startserver_functions.go
```diff
@@ -205,6 +205,14 @@ func CreateBlockchain(config *configuration.Config) (bc *blockchain.Blockchain,
...
 	}
+
+	listedTokens, err := bc.GetListedTokens()
+	if err != nil {
+		log.Panicf("cannot initiate listed token: %s", err)
+	}
+	bc.SetListedTokens(listedTokens)
+
+	// load all listed token indices instead of only configs tokens
 	err = bc.LoadAndSetTokenIndices(common.GetTokenAddressesList(tokens))
```
secmask: we need  to use `listedToken` here

http/settings_handler.go
```diff
@@ -417,7 +417,7 @@ func (s *Server) ensureInternalSetting(tokenUpdate common.TokenUpdate) error {
 			s.l.Errorw("failed to get listed token from blockchain", "error", err)
 			return err
 		}
-		s.listedTokens = listedTokens
+		s.blockchain.SetListedTokens(listedTokens)
```
secmask: this inside create pending change, we should put this in confirm step, around this block

``` golang
if hasInternal {
		if err = s.updateInternalTokensIndices(tokenUpdates); err != nil {
			httputil.ResponseFailure(c, httputil.WithReason(fmt.Sprintf("Can not update internal token indices (%+v)", err)))
			return
		}
		if err = s.metric.ConfirmTokenUpdateInfo(tarQty, pws, quadEq); err != nil {
			httputil.ResponseFailure(c, httputil.WithReason(fmt.Sprintf("Can not update metric data (%+v)", err)))
			return
		}
	}
```

PR #760, opened 2019-11-20
Update load and set token indices

halink0803:


secmask:
@halink0803 we should finish this one.

halink0803:
In core-v3, we have 2 services: core and reserve-settings, core in charge of set_rates(we only care about this now), which we have a blockchain object maintaining a list of tokens indices (index of tokens in list of listed token in pricing contract). reserve-settings is responsible for storing all settings (including tokens settings). 

In this task, we make sure that when we list a new token in reserve-settings, token indices in core is also updated and when we delist a token from dashboard, the set-rates function will set all delisted token rate to 0.

So, we have to test 2 case here: 1 - delist a token, set-rates and check if the delisted token rates be set to 0
2 - list a new token and make sure the token indices list is updated.

gateway/http/routes_proxy.go
```diff
@@ -41,6 +41,9 @@ func WithCoreEndpoint(coreEndpoint string) Option {
 
 		g.GET("/addresses", coreProxyMW)
 
+		g.PUT("/update-token-indice", coreProxyMW)
```
secmask: setting service call this through gateway? how do we deal with authentication?

halink0803: This is internal call, so in deployment this can point directly to core-endpoint, not sure if it is possible. Let me ask Thang.

secmask: so I think we don't need to add this endpoint to gateway, and setting point directly to core(internal traffic)

halink0803: @secmask I asked Thang, and he said it is possible to call directly from reserve-setting to core endpoint. I will removed this code from gateway as it is not neccessary.

PR #764, opened 2019-11-26
Update get activities

halink0803:


halink0803:
> LGTM.
> However, I don't recommend update go.mod or add vendor changes since it is not really neccessary

Yes, it is added automatically. I will look into it to find out a way to prevent that.

PR #770, opened 2019-11-27
Fix activity result currently return string nil

halink0803:


secmask:
is this need on v1 also?

PR #772, opened 2019-11-28
body should convert to string before log

secmask:
convert body to string when log, otherwise sentry will show []byte in base64 format


secmask:
> pls check len(body) before deciding if we should log the body or not

hm, check if it too long?

secmask:
> > pls check len(body) before deciding if we should log the body or not
> 
> hm, check if it too long?

updated.

PR #777, opened 2019-12-03
update internal network contract

halink0803:


secmask:
@halink0803 it conflict now

PR #778, opened 2019-12-03
Update internal network for v3

halink0803:


cmd/configuration/contract_addresses.go
```diff
@@ -38,7 +38,7 @@ var defaultAddressConfiguration = map[deployment.Deployment]common.ContractAddre
 		Reserve:         ethereum.HexToAddress("0x63825c174ab367968EC60f061753D3bbD36A0D8F"),
 		Wrapper:         ethereum.HexToAddress("0x6172AFC8c00c46E0D07ce3AF203828198194620a"),
 		Pricing:         ethereum.HexToAddress("0x798AbDA6Cc246D0EDbA912092A2a3dBd3d11191B"),
-		InternalNetwork: ethereum.HexToAddress("0x91a502C678605fbCe581eae053319747482276b9"),
```
secmask: update line 27 too?

halink0803: I have removed InternalNetwork config as we don't use it and nobody uses it.

PR #784, opened 2019-12-13
add base_volatility_spread to feed_configuration

NNagato:


data/storage/bolt.go
```diff
@@ -127,9 +129,41 @@ func NewBoltStorage(path string) (*BoltStorage, error) {
...
+			if dataBytes := b.Get([]byte(feed)); dataBytes == nil {
+				var fi feedInfo
+				if dataBytes := bdf.Get([]byte(feed)); dataBytes == nil {
+					fi.Enabled = true
+				}
+				dataJSON, err := json.Marshal(fi)
+				if err != nil {
+					return err
+				}
+				return b.Put([]byte(feed), dataJSON)
```
secmask: return here will abort migrate on first success feed?

data/storage/bolt.go
```diff
@@ -127,9 +129,41 @@ func NewBoltStorage(path string) (*BoltStorage, error) {
 		db: db,
 		l:  zap.S(),
 	}
+	if err = storage.migrateFeedConfiguration(); err != nil {
```
secmask: please add a todo to remove this migrate after it get deploy and run normally, as it no longer need.

data/storage/bolt.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-data/pull/784/commits/9e80c4ef75fe9cf7189b51b24142e8bd505935dc

data/storage/bolt.go
NNagato: updated in https://github.com/KyberNetwork/reserve-data/pull/784/commits/284b376f7a312e2c982235168fd460163575a936

PR #787, opened 2019-12-18
add api set feed setting

NNagato:


data/storage/bolt.go
```diff
@@ -287,6 +291,117 @@ func (bs *BoltStorage) GetFeedConfiguration() ([]common.FeedConfiguration, error
 	return results, err
 }
 
+func (bs *BoltStorage) StorePendingFeedSetting(value []byte) error {
+	var (
+		err         error
+		pendingData common.MapFeedSetting
```
secmask: we should verify that each feed in update request exist in AllFeed() list.

common/feeds.go
```diff
@@ -4,3 +4,9 @@ type FeedConfiguration struct {
 	Name    string `json:"name"`
 	Enabled bool   `json:"enabled"`
 }
+
+type FeedSetting struct {
```
secmask: public type should have comment.

common/feeds.go
```diff
@@ -4,3 +4,9 @@ type FeedConfiguration struct {
...
+	BaseVolatilitySpread float64 `json:"base_volatility_spread"`
+}
+
+type MapFeedSetting map[string]FeedSetting
```
secmask: public type should have comment.

common/feeds.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-data/pull/787/commits/7ee2f70c0bdd6a8b4aba5e17c9adc44560af6890

common/feeds.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-data/pull/787/commits/7ee2f70c0bdd6a8b4aba5e17c9adc44560af6890

data/storage/bolt.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-data/pull/787/commits/7ee2f70c0bdd6a8b4aba5e17c9adc44560af6890

NNagato:
> we should extract inner loop into function, so we don't have to check `if index == len(allFeeds)-1 {`

pls take a look, I updated in https://github.com/KyberNetwork/reserve-data/pull/787/commits/32befd4fa120f6b8ca4e1c4b0fbec89e321cd6f9

PR #790, opened 2019-12-26
move feed configuration to reserve setting

NNagato:


NNagato:
> > Delete key
> 
> fix copy&paste please

fixed in https://github.com/KyberNetwork/reserve-data/pull/790/commits/5f013543299944d6bfdc73662d854a67965952ce

PR #795, opened 2020-01-10
log tx hash when fetch tx status v1

NNagato:


data/fetcher/fetcher.go
```diff
@@ -372,6 +374,7 @@ func (f *Fetcher) FetchStatusFromBlockchain(pendings []common.ActivityRecord) (m
 			}
 			status, blockNum, err = f.blockchain.TxStatus(tx)
 			if err != nil {
+				f.l.Warnw("cannot get tx status", "tx hash", txStr)
 				return result, fmt.Errorf("TX_STATUS: ERROR Getting tx status failed: %s", err)
```
secmask: we can include `tx` in `fmt.Errorf....` as caller will log this error out, this way we don't have 2 log line for this error.

NNagato: pr updated in https://github.com/KyberNetwork/reserve-data/pull/795/commits/97249f74c186fd472e15d00a0a1e47c91b891d25

PR #796, opened 2020-01-10
log tx hash when fetch tx status

NNagato:


halink0803:
@secmask This task long ago and I think it can be merged.

secmask:
ok

PR #801, opened 2020-01-13
Fix uneeded log

secmask:


data/fetcher/fetcher.go
```diff
@@ -500,10 +504,7 @@ func (f *Fetcher) updateActivitywithExchangeStatus(activity *common.ActivityReco
 		activity.ExchangeStatus = activityStatus.ExchangeStatus
 	}
 
-	resultTx, ok := activity.Result["tx"].(string)
-	if !ok {
-		f.l.Warnw("activity.Result[tx] cannot be asserted to string type", "tx", fmt.Sprintf("[%+v]", activity.Result["tx"]))
-	} else if ok && resultTx == "" {
+	if resultTx, ok := activity.Result["tx"].(string); ok && resultTx == "" {
```
halink0803: here we ignore when the activity.Result["tx"] cannot be cast to string? I think there should be a comment reason for that.

secmask: base on what I check in log, it failed to cast to string because it wasn't set (is nil), and this is safe to ignore.
this change should have the same affect, except don't log because it is not useful.

PR #802, opened 2020-01-14
replace geminiBTC feed by binanceBTC

secmask:


secmask:
@ductm54 cc

world/endpoint.go
```diff
@@ -14,7 +14,8 @@ var (
 		"Gemini",
 
 		"CoinbaseBTC",
-		"GeminiBTC",
+		"GeminiBTC", // this no longer use and replace by BinanceBTC below.
```
halink0803: Then why don't we remove it from the list?

common/config/config.go
```diff
@@ -25,6 +25,7 @@ type WorldEndpoints struct {
 
 	CoinbaseBTC SiteConfig `json:"coinbase_btc"`
 	GeminiBTC   SiteConfig `json:"gemini_btc"`
```
halink0803: Should it be removed as it is no longer used?

world/endpoint.go
secmask: because we still have it in feed configuration(and we dont have a way to remove it now), so we keep it in all-feeds list to not surprise frontend team.  say, we dont use GeminiBTC as in using feed, but it still an option later.

common/config/config.go
secmask: ok, removed

PR #805, opened 2020-01-16
don't warn if got nil tx in activity

secmask:


secmask:
an other missing place.

PR #806, opened 2020-01-16
set default time range get-all-rates

NNagato:


http/server.go
```diff
@@ -168,7 +169,10 @@ func (s *Server) GetRates(c *gin.Context) {
 	fromTime, _ := strconv.ParseUint(c.Query("fromTime"), 10, 64)
 	toTime, _ := strconv.ParseUint(c.Query("toTime"), 10, 64)
 	if toTime == 0 {
-		toTime = maxTimespot
+		toTime = common.TimeToMillis(time.Now())
```
secmask: should we also have a limit for max time-range?

NNagato: as we discussed, to avoid breaking other process, maybe we will add limit max time-range latter

PR #808, opened 2020-01-17
move secret config to separate file

secmask:


secmask:
@halink0803 review this please.

PR #811, opened 2020-01-20
move hard code to config file

NNagato:


world/world.go
```diff
@@ -151,7 +151,7 @@ func (tw *TheWorld) GetGoldInfo() (common.GoldData, error) {
 }
 
 //NewTheWorld return new world instance
-func NewTheWorld(dpl deployment.Deployment, keyfile string) (*TheWorld, error) {
+func NewTheWorld(dpl deployment.Deployment, configFile string) (*TheWorld, error) {
```
secmask: we no longer need `SimulatedEndpoint`, as all url now load from config file, `dpl` param can remove also

cmd/configuration/config_files.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// NewSecretConfigCliFlag returns the cli flag to configure secret config file flag.
+func NewSecretConfigCliFlag() []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			Name:   secretConfigFileFlag,
+			Usage:  "location of secret config file",
+			EnvVar: "SECRET_CONFIG_FILE",
+			Value:  filepath.Join(common.CmdDirLocation(), "secret_config.json"),
```
secmask: we dont need to join `common.CmdDirLocation()`, as config file should be passed as service parameter.

cmd/configuration/config_files.go
```diff
@@ -0,0 +1,42 @@
...
+		},
+		cli.StringFlag{
```
secmask: we dont need to join common.CmdDirLocation(), as config file should be passed as service parameter.

NNagato: changed as request

cmd/configuration/config_files.go
NNagato: changed as request

world/world.go
NNagato: changed as request

world/endpoint.go
```diff
@@ -87,7 +89,7 @@ type Endpoint interface {
 
 // RealEndpoint return real endpoint
 type RealEndpoint struct {
-	OneForgeKey string `json:"oneforge"`
+	EPS common.WorldEndpoints `json:"eps"`
```
halink0803: Using "EPS" seems confusing, should use understandable name like ConfigEndpoints, AllEndpoints, etc.

NNagato: PR updated in https://github.com/KyberNetwork/reserve-data/pull/811/commits/e2e8da039841591f996060b2fdfe23bdee90ff0f

secmask:
@NNagato I update the issue with one more task, please check it out.

NNagato:
> @NNagato I update the issue with one more task, please check it out.

PR updated, please help to review

PR #816, opened 2020-02-04
Binance 2

halink0803:


exchange/huobi/blockchain/huobi_blockchain.go
```diff
@@ -48,6 +49,8 @@ func NewBlockchain(
 
 	base.MustRegisterOperator(HuobiOP, blockchain.NewOperator(signer, nonce))
 
+	log.Printf("register huobi intermediate operator: %s", signer.GetAddress().Hex())
```
secmask: is it necessary to print this?

halink0803: log removed.

PR #822, opened 2020-02-07
add support coinbase, currently only fetch orderbook

secmask:


halink0803:
I think you could add more logger when there is error, it could make us easier to detect error if having any. Generally, it looks good to me.

PR #838, opened 2020-02-18
Rolldata fix metric10k records

secmask:


data/storage/bolt_migrate.go
```diff
@@ -93,17 +99,42 @@ func (bs *BoltStorage) Migrate(newbs *BoltStorage) (err error) {
...
+	}
 
 	// other buckets
 	errU := newbs.db.Update(func(newTX *bolt.Tx) error {
 		errV := bs.db.View(func(tx *bolt.Tx) error {
+
+			bs.l.Info("process metric info")
+			mtricBucket := tx.Bucket([]byte(metricBucket))
+			mtricBucketNew := newTX.Bucket([]byte(metricBucket))
+			maxMetricCopyMax := 10000
```
NNagato: should we make this value as a flag? and if not I think we need a better name for it like maxMetricRecordCopy,...

secmask: we mostly don't have a reason to change its value.

PR #847, opened 2020-02-24
support enable disable feed without confirm v3

NNagato:


secmask:
@ductm54 @edwardEvans094 cc

PR #848, opened 2020-02-24
Update open orders

halink0803:


common/exchange.go
```diff
@@ -41,6 +41,9 @@ type Exchange interface {
 	Address(asset common.Asset) (address ethereum.Address, supported bool)
 	Withdraw(asset common.Asset, amount *big.Int, address ethereum.Address) (string, error)
 	Trade(tradeType string, pair common.TradingPairSymbols, rate, amount float64) (id string, done, remaining float64, finished bool, err error)
+
+	// OpenOrders return open orders from exchange
+	OpenOrders(pair common.TradingPairSymbols) (orders []Order, err error)
```
secmask: Coinbase exchange need to add this also to satisfy Exchange interface.

common/types.go
```diff
@@ -450,20 +450,21 @@ type AllBalanceResponse struct {
 	Data       map[string]BalanceResponse
 }
 
+// Order accross multiple exchanges
 type Order struct {
-	ID          string // standard id across multiple exchanges
-	Base        string
+	ID          string `json:"ID,omitempty"` // standard id across multiple exchanges
+	Base        string `json:"Base"`
 	Quote       string
```
secmask: why some fields does not have json tag?

halink0803: this is a legacy declare, it used to return the json like its name (I added json tag only for omit some not important fields). Anw, I will add json tag for all of them.

common/exchange.go
halink0803: Added it.

exchange/binance.go
```diff
@@ -399,6 +399,46 @@ func (bn *Binance) OrderStatus(id string, base, quote string) (string, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		result = append(result, common.Order{
+			OrderID: strconv.FormatUint(order.OrderID, 10),
+			Side:    order.Side,
+			Type:    order.Type,
+			OrigQty: originalQty,
+			Price:   price,
+			Base:    pair.BaseSymbol,
```
secmask: please comment detail on the way we fill symbol here.

PR #870, opened 2020-03-05
Update docs and cancel order to accept multiple order ids

halink0803:


core/reserve_core.go
```diff
@@ -52,17 +52,22 @@ func timebasedID(id string) common.ActivityID {
...
+		activity, err := rc.activityStorage.GetActivity(exchange.ID(), orderID)
+		if err != nil {
+			return err
+		}
+		if activity.Action != common.ActionTrade {
+			return errors.New("this is not an order activity so cannot cancel")
+		}
+		base := activity.Params.Base
+		quote := activity.Params.Quote
+		if err := exchange.CancelOrder(orderID, base, quote); err != nil {
```
secmask: this would stop on first fail, I think it's safe to continue, then we can have a list of result for each orderID

core/reserve_core.go
```diff
@@ -52,17 +52,26 @@ func timebasedID(id string) common.ActivityID {
...
-		return err
-	}
-	if activity.Action != common.ActionTrade {
-		return errors.New("this is not an order activity so cannot cancel")
+func (rc ReserveCore) CancelOrder(orderIDs []string, exchange common.Exchange) map[string]string {
+	result := make(map[string]string)
+	for _, orderID := range orderIDs {
+		activity, err := rc.activityStorage.GetActivity(exchange.ID(), orderID)
+		if err != nil {
+			return result
```
secmask: we can add a result entry here also instead of return

core/reserve_core.go
halink0803: Updated.

core/reserve_core.go
halink0803: change as requested.

secmask:
LGTM

PR #890, opened 2020-03-16
Add test for set feed weight

halink0803:


secmask:
please also add test on setFeedConfiguration for support/unsupport base on set_rate

halink0803:
> please also add test on setFeedConfiguration for support/unsupport base on set_rate
as I check the setFeedConfiguration is storage function which does not depend on the value of set_rate for supported or unsupported. I have added the test case for func checkSetFeedConfiguration in http package but it seems does not depend on set_rate value also. Anw, I will try to add more test cases on http also.

https://github.com/KyberNetwork/reserve-data/blob/8ee20dc2c15c520833b7a792e2ad6a139e16c539/reservesetting/http/setting_change_test.go#L94

secmask:
> > please also add test on setFeedConfiguration for support/unsupport base on set_rate
> > as I check the setFeedConfiguration is storage function which does not depend on the value of set_rate for supported or unsupported. I have added the test case for func checkSetFeedConfiguration in http package but it seems does not depend on set_rate value also. Anw, I will try to add more test cases on http also.
> 
> https://github.com/KyberNetwork/reserve-data/blob/8ee20dc2c15c520833b7a792e2ad6a139e16c539/reservesetting/http/setting_change_test.go#L94

@halink0803 
yes, but it's different now that we split feed into category,  feedConfiguration now a pair of `feedName-setRate`. I think we can improve the checking here https://github.com/KyberNetwork/reserve-data/blob/8ee20dc2c15c520833b7a792e2ad6a139e16c539/reservesetting/http/setting_change.go#L595
we will check for one feed type (usd,gold,btc) base on setRate value, and give more specific error message.

reservesetting/http/setting_change.go
```diff
@@ -594,12 +594,30 @@ func (s *Server) checkUpdateExchangeParams(updateExchangeEntry common.UpdateExch
 
 func (s *Server) checkSetFeedConfigurationParams(setFeedConfigurationEntry common.SetFeedConfigurationEntry) error {
 	if _, ok := feed.AllFeeds().BTC[setFeedConfigurationEntry.Name]; ok {
+		if setFeedConfigurationEntry.SetRate != common.BTCFeed {
```
secmask: we can have a `switch(setFeedConfigurationEntry.SetRate)` would be simpler?

halink0803: I have changed as suggested.

PR #891, opened 2020-03-18
5590 using dynamic gasprice

secmask:


data/storage/bolt.go
```diff
@@ -1022,13 +1023,13 @@ func (bs *BoltStorage) RemoveStalePendingActivities(tx *bolt.Tx, stales []common
 	return nil
 }
 
-//PendingSetRate return pending set rate activity
-func (bs *BoltStorage) PendingSetRate(minedNonce uint64) (*common.ActivityRecord, uint64, error) {
+//PendingActivityForAction return pending set rate activity
```
halink0803: Here the pending activity is not only for setrate but for deposit also.

data/fetcher/blockchain.go
```diff
@@ -12,5 +14,5 @@ type Blockchain interface {
 	FetchRates(atBlock uint64, currentBlock uint64) (common.AllRateEntry, error)
 	TxStatus(tx ethereum.Hash) (string, uint64, error)
 	CurrentBlock() (uint64, error)
-	SetRateMinedNonce() (uint64, error)
+	blockchain.MinedNonePicker
```
halink0803: It should be "Nonce" instead of "None"?

core/reserve_core.go
```diff
@@ -273,6 +259,70 @@ func (rc ReserveCore) Deposit(
...
+	address, supported := exchange.Address(token)
+	if !supported {
+		return nil, fmt.Errorf("exchange %s doesn't support token %s", exchange.ID(), token.ID)
+	}
+	found, err := rc.activityStorage.HasPendingDeposit(token, exchange)
+	if err != nil {
+		return nil, err
+	}
+	if found {
+		return nil, fmt.Errorf("there is a pending %s deposit to %s currently, please try again", token.ID, exchange.ID())
```
halink0803: We return when we see a pending deposit here. Then we cannot replace it later. The comment below seems not correct

core/reserve_core.go
```diff
@@ -273,6 +259,70 @@ func (rc ReserveCore) Deposit(
...
+	}
+
+	if err = sanityCheckAmount(exchange, token, amount); err != nil {
+		return nil, err
+	}
+
+	// if there is a pending deposit tx, we replace it
```
halink0803: We cannot reach here if there is a pending deposit.

data/fetcher/blockchain.go
secmask: oh, that's right

blockchain/blockchain.go
```diff
@@ -61,10 +56,11 @@ type Blockchain struct {
 	listedTokens []ethereum.Address
 	mu           sync.RWMutex
 
-	localSetRateNonce     uint64
-	setRateNonceTimestamp uint64
-	setting               Setting
-	l                     *zap.SugaredLogger
+	localSetRateNonce, localDepositNone          uint64
```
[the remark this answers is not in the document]
secmask: good catch

core/reserve_core.go
secmask: I removed check pendingDeposit part

data/storage/bolt.go
secmask: fixed

data/fetcher/blockchain.go
secmask: fixed

blockchain/blockchain.go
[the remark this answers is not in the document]
secmask: fixed

PR #893, opened 2020-03-19
update usd,btc,gold feed name in response

secmask:


cmd/example_config.json
```diff
@@ -70,7 +70,7 @@
       "url": "https://api.pro.coinbase.com/products/eth-dai/ticker"
     },
     "coinbase_dai_10000": {
-      "url": "http://localhost:8003/feed/CoinbaseETHDAI10000"
+      "url": "http://localhost:8043/feed/CoinbaseETHDAI10000"
```
halink0803: Best practice we do not show port number in public?

secmask: just a random number, like our http listen port.

PR #904, opened 2020-03-25
Update api docs

halink0803:
Minor update

### DIFF
--- a/http/server.go
+++ b/http/server.go
@@ -322,7 +322,7 @@ func (s *Server) CancelOrder(c *gin.Context) {
 		return
 	}
 	s.l.Infow("Cancel order", "id", request.OrderIDs, "from", exchange.ID().String())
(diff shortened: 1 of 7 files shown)


http/server.go
```diff
@@ -532,7 +532,7 @@ func (s *Server) register() {
 		g.GET("/immediate-pending-activities", s.ImmediatePendingActivities)
 
 		g.GET("/open-orders", s.OpenOrders)
-		g.POST("/cancelorder", s.CancelOrder)
+		g.POST("/cancel-order", s.CancelOrder)
```
secmask: is it `cancel-order` or `cancel-orders`?

halink0803: change to cancel-orders, make more sense.

halink0803:
Change as requested also fix null response.

PR #905, opened 2020-03-25
fix not set gasconfig

secmask:


blockchain/blockchain.go
```diff
@@ -77,15 +77,19 @@ func (b *Blockchain) StandardGasPrice() float64 {
 	if b.gasConfig.PreferUseGasStation {
 		gss, err := b.gsClient.ETHGas()
 		if err == nil { // TODO: we can make decision to use gss.Fast to other if some one ask.
-			return gss.Fast / 10.0 // gas return by gas station is *10, so we need to divide it here
+			res := gss.Fast / 10.0
+			b.l.Infow("use gas_price from gasstation", "value", res)
+			return res // gas return by gas station is *10, so we need to divide it here
```
halink0803: it seems the comment is not in correct place

PR #908, opened 2020-03-26
update docs for open and cancel order

halink0803:


api_docs.md
```diff
@@ -6,34 +6,39 @@
...
 ```json
-    {
-        "success": true,
-        "data": [
-            "OrderID": "123132",
-            "Price": 0.43432,
+{
+    "success": true,
+    "data": {
+        "1": [
```
ductm54: in this example it should be binance or something similar. using "1" is ambigous here

halink0803: Yep, this example seems not correct, let me fix it.

PR #915, opened 2020-03-28
support use gasstation and max gas from contract

NNagato:


common/types.go
```diff
@@ -762,6 +762,12 @@ type FetcherDelay struct {
 	TradeHistory  HumanDuration `json:"trade_history"`
 }
 
+// GasConfig ...
+type GasConfig struct {
+	PreferUseGasStation     bool  `json:"prefer_use_gas_station"`
```
secmask: LGTM, a small change to add example config for this new entry

NNagato: > LGTM, a small change to add example config for this new entry

added in https://github.com/KyberNetwork/reserve-data/pull/915/commits/d8178cbe5849da44c3b67d1eeba97525333ad303

secmask: I thought it look like 
```
"gas_config": {
    "prefer_use_gas_station": true,
    "fetch_max_gas_cache_seconds": 300
  }
```

NNagato: fixed in https://github.com/KyberNetwork/reserve-data/pull/915/commits/76a3098c07bc34a1bf4f93e23522d80f8d1ce293

NNagato: > I thought it look like
> 
> ```
> "gas_config": {
>     "prefer_use_gas_station": true,
>     "fetch_max_gas_cache_seconds": 300
>   }
> ```

fixed in https://github.com/KyberNetwork/reserve-data/pull/915/commits/76a3098c07bc34a1bf4f93e23522d80f8d1ce293

PR #947, opened 2020-04-24
Update binance api v3

halink0803:
- Update `quotePrecision` to `quoteAssetPrecision`
- Add api DELETE `api/v3/openOrders`

### DIFF
--- a/exchange/binance/binance_endpoint.go
+++ b/exchange/binance/binance_endpoint.go
@@ -293,6 +293,26 @@ func (ep *Endpoint) CancelOrder(symbol string, id uint64) (exchange.Binacancel, 
 	return result, err
 }
 
+// CancelAllOrders cancel all open order of an symbols
+func (ep *Endpoint) CancelAllOrders(symbol string) ([]exchange.Binaorder, error) {
+	var result []exchange.Binaorder
+	respBody, err := ep.GetResponse(
+		"DELETE",
+		ep.interf.AuthenticatedEndpoint()+"/api/v3/openOrders",
+		map[string]string{
+			"symbol": symbol,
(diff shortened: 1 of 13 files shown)


exchange/binance_format.go
```diff
@@ -67,7 +67,7 @@ type FilterLimit struct {
 type BinanceSymbol struct {
 	Symbol             string        `json:"symbol"`
 	BaseAssetPrecision int           `json:"baseAssetPrecision"`
-	QuotePrecision     int           `json:"quotePrecision"`
+	QuotePrecision     int           `json:"quoteAssetPrecision"`
```
secmask, line 70: can we rename `QuotePrecision` to `QuoteAssetPrecision` too?

exchange/binance/binance_endpoint.go
```diff
@@ -293,6 +293,26 @@ func (ep *Endpoint) CancelOrder(symbol string, id uint64) (exchange.Binacancel,
 	return result, err
 }
 
+// CancelAllOrders cancel all open order of an symbols
+func (ep *Endpoint) CancelAllOrders(symbol string) ([]exchange.Binaorder, error) {
+	var result []exchange.Binaorder
+	respBody, err := ep.GetResponse(
+		"DELETE",
```
secmask, line 300: there is `http.MethodDelete` constant for method.

exchange/binance_format.go
halink0803: I have updated it.

exchange/binance/binance_endpoint.go
halink0803: I have changed it.

PR #948, opened 2020-04-27
Update binance api

halink0803:


secmask:
@ductm54 we update due bianance changed their API, please test this.

PR #964, opened 2020-05-11
Update check token indice for update asset entry

halink0803:


reservesetting/http/setting_change.go
```diff
@@ -507,39 +514,46 @@ func checkFeedWeight(setrate *common.SetRate, feedWeight *common.FeedWeight) err
 	return nil
 }
 
+func (s *Server) checkTokenIndices(tokenAddress ethereum.Address) error {
+	if s.coreEndpoint != "" { // check to by pass test as local test does not need this
+		// check token indice in core
+		endpoint := fmt.Sprintf("%s/v3/check-token-indice?address=%s", s.coreEndpoint, tokenAddress.Hex())
```
secmask, line 520: look like we can move this logic into a client object.

PR #976, opened 2020-05-27
fix feed number

NNagato:


cmd/example.json
```diff
@@ -129,7 +129,7 @@
       "url": "http://localhost:8003/feed/BinanceETHPAX5000"
     },
     "binance_busd_100000": {
-      "url": "http://localhost:8003/feed/BinanceETHBUSD100000"
```
secmask, line 132: update this too `  "binance_busd_100000"` ?

NNagato: fixed in https://github.com/KyberNetwork/reserve-data/pull/976/commits/b25562525e2ec40e043d4b4a293a18abf91610d2

PR #999, opened 2020-07-17
Update binance get trade history query

halink0803:


exchange/binance/storage/binance_postgres_test.go
```diff
@@ -49,7 +53,7 @@ func Test_BinancePostgres(t *testing.T) {
 	var tradeHistory common.ExchangeTradeHistory
 	fromTime := uint64(1528934400000)
 	toTime := uint64(1529020800000)
-	tradeHistory, err = storage.GetTradeHistory(fromTime, toTime)
+	tradeHistory, err = storage.GetTradeHistory(1, fromTime, toTime)
```
secmask, line 56: let use `Binance` instead of `1`, I remember it defined some where.

exchange/binance/storage/binance_postgres_test.go
```diff
@@ -10,15 +10,19 @@ import (
 	"github.com/KyberNetwork/reserve-data/common"
 	"github.com/KyberNetwork/reserve-data/common/testutil"
 	"github.com/KyberNetwork/reserve-data/exchange"
+	"github.com/KyberNetwork/reserve-data/reservesetting/storage/postgres"
 )
 
 func Test_BinancePostgres(t *testing.T) {
+	t.Skip() // skip as we do not have a way to update database yet
```
secmask, line 17: we want to skip this?

halink0803: No, I am thinking about how we init data for test.

exchange/binance/storage/binance_postgres_test.go
halink0803: Yes, it makes sense.

PR #1000, opened 2020-07-21
modify filed name in asset

NNagato:


halink0803:
@NNagato Could you please ask Spyros and Edward about this changes. I thought the key cause the confusion is id (under exchanges), not the exchange itself. I think this change might still make confuse. Otherwise, it is good to go.

NNagato:
> @NNagato Could you please ask Spyros and Edward about this changes. I thought the key cause the confusion is id (under exchanges), not the exchanges itself. I think this change might still make confuse. Otherwise, it is good to go.

Ya, I will confirm with Spyros and Edward. Currently, the `id` make confuse cuz we don't know what does `id` in exchanges dict mean. But the `exchanges` field actually is `asset exchanges` so I think change its name is a better solution.

secmask:
@edwardEvans094 could you have a look on these changes? I will merge this you ready

PR #1008, opened 2020-07-24
prepare statements after migrate

secmask:


exchange/binance/storage/binance_postgres.go
```diff
@@ -68,7 +66,7 @@ func NewPostgresStorage(db *sqlx.DB, migrationFolderPath, databaseName string) (
 			return storage, err
 		}
 	}
-
+	err = storage.prepareStmts()
```
NNagato, line 69: ah, can we refactor this func by removing err in var scope and just return storage, storage.prepareStmts()?

PR #1019, opened 2020-07-29
update docs for token rate trigger and setrate

secmask:


apidocs/source/includes/settings/_rate_trigger.md
```diff
@@ -1,8 +1,8 @@
-# Rate trigger period
+# Rate trigger
 
 ## Set rate trigger period
```
NNagato, line 3: should we change it to "Set rate trigger" too?

secmask: nop, setrate trigger period is correct, it's child element in Rate Trigger page.

PR #1020, opened 2020-07-30
ignore not listed token from get rate

secmask:


secmask:
> This error really confuse me as it is not happened in v1. But it is better checked. LGTM only need to fix golangci-lint

I checked, it actually cause by adding a new asset(with set_rate) but not list it yet.

PR #1027, opened 2020-08-05
handle 429 binance status code

secmask:


exchange/transport_limiter.go
```diff
@@ -32,18 +32,35 @@ func NewTransportRateLimiter(c *http.Client) *TransportRateLimiter {
 		l: zap.S(),
 	}
 }
+
+func nextMin() time.Time {
+	now := time.Now()
+	nextMin := now.Add(time.Second * time.Duration(60-now.Second()))
```
halink0803, line 38: Why this is not now.Add(time.Minute)? Please add comment to explain if it is special handle.

secmask: yeah, it would be simpler

PR #1029, opened 2020-08-06
use cex-account-data to get account info

NNagato:


cmd/example_secret.json
```diff
@@ -1,6 +1,10 @@
 {
+  "binance_account_id": "b1",
```
secmask, line 2: could you move this line down, near by its relate entry for easy tracking

NNagato: > could you move this line down, near by its relate entry for easy tracking

I think it's already in the right place with cluster [account id, api key, sec key] they belong to one Binance account

secmask: ah yep

PR #1121, opened 2020-10-07
deposit with different asset or exchange now use increase nonce, fix …

secmask:
…problem where node reject use same nonce


halink0803 (approved):
LGTM

PR #1125, opened 2020-10-09
create missing index, trade history not necessary to ref trading_pair…

secmask:
… as trading_pair can be delete


halink0803 (approved):
LGTM

PR #1176, opened 2020-11-15
adjust formular calculate new price for override setRate

secmask:
I propose a change the way we calculate new price for override nonce
details: 
current method: of we count number of pending tx and use as step increasing gasPrice [initPrice -> maxPrice]
new method: we stick with gasPrice from gas-service, but it should >10% compare to last recent pending price(so node won't complain about underprice, I guess)

this is because our maxPrice is pretty high now(1000 gwei), so current method can use unnecessary high price.


halink0803 (approved):

halink0803:
LGTM

NNagato (approved):

PR #1192, opened 2020-11-23
Add reserve settings api to get exchange trading pairs

halink0803:


reservesetting/http/trading_pair.go
```diff
@@ -30,6 +30,27 @@ func (s *Server) getTradingPair(c *gin.Context) {
 	httputil.ResponseSuccess(c, httputil.WithData(result))
 }
 
+type getTradingPairParam struct {
+	ID rtypes.ExchangeID `form:"id" binding:"required"`
```
secmask, line 34: as it map to /trading-pair, I would suggest to name this request param to `exchange` to avoid confuse with `trading pair id`

secmask: and also need to update api docs

halink0803: Updated the field and api docs

secmask (approved):

PR #1207, opened 2020-11-27
Add new method for authentication and support multiple approve

NNagato:


secmask:
@halink0803 could you help to review this?

halink0803:
@secmask yes, I can.

reservesetting/http/setting_change.go
```diff
@@ -304,14 +304,54 @@ func (s *Server) rejectSettingChange(c *gin.Context) {
...
+		input struct {
+			ID uint64 `uri:"id" binding:"required"`
+		}
+	)
+	if err := c.ShouldBindUri(&input); err != nil {
+		httputil.ResponseFailure(c, httputil.WithError(err))
+		return
+	}
+	keyID := s.getKeyIDFromContext(c)
+	if err := s.storage.DispproveSettingChange(keyID, input.ID); err != nil {
```
secmask, line 322: there's a typo in `DisapproveSettingChange`

gateway/cmd/config.json.example
```diff
@@ -0,0 +1,26 @@
+{
+  "read_role": [
+    {
+      "AccessKeyID": "read_access_key_id",
```
secmask, line 4: it's not consistent to use uppercase here, we can redefine KeyPair struct with json tag, I would suggest also add optional `name` field for keyPair so Gary can easy to remember who this key belong to.

reservesetting/http/setting_change.go
NNagato: fixed in https://github.com/KyberNetwork/reserve-data/pull/1207/commits/15adffd235ddfc95d487636c639e79750465eee4

gateway/cmd/config.json.example
NNagato: updated in https://github.com/KyberNetwork/reserve-data/pull/1207/commits/15adffd235ddfc95d487636c639e79750465eee4

secmask (approved):

PR #1210, opened 2020-11-30
refactor fetch exchange data

secmask:
- refactor fetch exchange data in the way easier to limit number of concurrent request
- move price(orderbook data) to separate table, so we can partition it daily and clean up old data.
- fetch_data table no longer compress data, read operation will detect and decompress old data on needed.


exchange/binance.go
```diff
@@ -180,71 +175,62 @@ func (bn *Binance) FetchOnePairData(
...
-		exPrice, ok := value.(common.ExchangePrice)
-		if !ok {
-			err = fmt.Errorf("value (%v) cannot be asserted to ExchangePrice", value)
-			return false
-		}
-		result[tokenPairID] = exPrice
-		return true
-	})
+	close(jobs)
+	concurrentFactor := 8
```
halink0803, line 220: this is fixed number, should it be a const?

halink0803 (approved):
LGTM. One question still.

exchange/binance.go
secmask: okay, I make it const

halink0803 (approved):
LGTM

PR #1216, opened 2020-12-01
let fetcher update order status instead of check it in Trade(), order…

secmask:
… not exists will has longer time window to verify, as binance can take long time to show up order in spike time


halink0803 (approved):
LGTM. The logic seems more simple now.

PR #1220, opened 2020-12-10
correct activity state on for failed tx

secmask:
this MR addresses below issues:
- fix speedupDeposit does not always increase gasPrice, now it use same calculate function with override nonce for setRate
- override nonce activity now show error message as `replace by tx ...`
- extend monitor lost tx by 5 minutes, to make sure it actually lost, in mean time, if we found an other tx with same nonce and mined, it will help confirm replaced tx early.


halink0803 (approved):
LGTM

PR #1260, opened 2021-02-18
add api return all asset withdraw status

NNagato:


http/server.go
```diff
@@ -614,6 +614,42 @@ func (s *Server) checkTokenIndice(c *gin.Context) {
...
+func (s *Server) getAllAssetWithdrawStatus(c *gin.Context) {
+	exchange := c.Param("exchange")
+	switch exchange {
+	case rtypes.Binance.String(), rtypes.Binance2.String():
+		data, err := s.binanceMainAccount.GetAllAssetWithdrawStatus()
+		if err != nil {
+			httputil.ResponseFailure(c, httputil.WithError(err))
+			return
+		}
+		httputil.ResponseSuccess(c, httputil.WithData(data))
```
secmask, line 626: result should be assetID instead of token string.

secmask (changes requested):

http/server.go
NNagato: changed in https://github.com/KyberNetwork/reserve-data/pull/1260/commits/ba0c91a5979f18a9bf3a5cee70628cab6cb56046

secmask:
looks good, but I would like to take 2 changes
- instead of return a list of asset withdraw status, we return a list of disabled withdraw asset, so any asset that not appear in our list will be assume as enabled to withdraw(in case this API failed somehow, will assume as withdraw allowed)
- update gateway for this API needed?

NNagato:
> looks good, but I would like to take 2 changes
> 
> * instead of return a list of asset withdraw status, we return a list of disabled withdraw asset, so any asset that not appear in our list will be assume as enabled to withdraw(in case this API failed somehow, will assume as withdraw allowed)
> * update gateway for this API needed?

I made the change in this commit https://github.com/KyberNetwork/reserve-data/pull/1260/commits/7d173782844232b244e59a3dcdc218f610758c66

secmask (approved):

PR #1296, opened 2021-03-30
fix minors with setrate / cancel-setrate

secmask:
- fix set block on set rate
- fix set pricing op set cancel set toAddress


halink0803 (approved):
LGTM

PR #1321, opened 2021-05-05
fix override setRate

secmask:
- fill TxTime for SetRate for override tx can work correctly
- - fix generate override tx activity id for setRate


halink0803 (approved):
LGTM

PR #1323, opened 2021-05-05
Update binance deposit intermediator

halink0803:


exchange/binance.go
```diff
@@ -354,8 +370,167 @@ func (bn *Binance) GetTradeHistory(fromTime, toTime uint64) (common.ExchangeTrad
...
+		if err == sql.ErrNoRows {
+			err := bn.process1stTx(id, asset, txHash, network, amount)
+			return common.ExchangeStatusNA, err
+		}
+
+		var (
+			data common.TXEntry
+		)
+		// if there is tx2Entry, check it blockchain status and handle the status accordingly:
+		miningStatus, _, err := bn.binBlockchain.TxStatus(ethereum.HexToHash(txHash))
```
secmask, line 480: should assign `txHash = txEntry.Hash` before bellow block can be run correctly.

cmd/migrations/000021_add_binance_intermediate_tx_up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE IF NOT EXISTS "binance_intermediate_tx"(
```
secmask, line 1: no need for `IF NOT EXISTS` with migration.

exchange/binance.go
```diff
@@ -530,16 +705,22 @@ func (bn *Binance) OpenOrders(pair commonv3.TradingPairSymbols) ([]common.Order,
 }
 
 // NewBinance init new binance instance
-func NewBinance(id rtypes.ExchangeID, interf BinanceInterface, storage BinanceStorage, sr storage.Interface) (*Binance, error) {
+func NewBinance(id rtypes.ExchangeID, interf BinanceInterface,
+	storage BinanceStorage, sr storage.Interface, blockchain *blockchain.BaseBlockchain, signer blockchain.Signer, nonce blockchain.NonceCorpus) (*Binance, error) {
+	bc, err := binblockchain.NewBlockchain(blockchain, signer, nonce, BinanceOP)
```
secmask, line 710: binBlockchain can be init out side and pass into this constructor

secmask (changes requested):

exchange/binance.go
```diff
@@ -23,6 +28,8 @@ import (
 
 const (
 	binanceEpsilon float64 = 0.0000001 // 10e-7
+	BSCNetwork             = "BSC"
```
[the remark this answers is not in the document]
halink0803: I use this constant instead of fixed value in other places.

exchange/binance.go
```diff
@@ -354,8 +370,167 @@ func (bn *Binance) GetTradeHistory(fromTime, toTime uint64) (common.ExchangeTrad
...
 func (bn *Binance) DepositStatus(id common.ActivityID, txHash string, assetID rtypes.AssetID, amount float64, timepoint uint64) (string, error) {
+	network := bcnetwork.GetPreConfig().Network
+	asset, err := bn.sr.GetAsset(assetID)
+	if err != nil {
+		return common.ExchangeStatusNA, err
+	}
+	if network == BSCNetwork && asset.Symbol == "BNB" { // this only apply for deposit BNB on BSC
+		// find second tx
+		txEntry, err := bn.storage.GetPendingIntermediateTx(id)
+		if err != nil && err != sql.ErrNoRows {
```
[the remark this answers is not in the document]
halink0803: What should I do with this? I don't know the better approach here.

exchange/binance.go
halink0803: assigned txHash before this block

exchange/binance.go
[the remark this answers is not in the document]
secmask: we can define our error const/var, like ErrNotFound

exchange/binance.go
```diff
@@ -354,8 +370,168 @@ func (bn *Binance) GetTradeHistory(fromTime, toTime uint64) (common.ExchangeTrad
...
+	tx, err := bc.BuildSendETHTx(opts, exchangeAddress)
+	if err != nil {
+		return nil, err
+	}
+	return bc.SignAndBroadcast(tx, blockchain.BinanceOP)
+}
+
+// Send2ndTransaction send the second transaction
+func (bn *Binance) Send2ndTransaction(amount float64, asset commonv3.Asset, exchangeAddress ethereum.Address) (*types.Transaction, error) {
+	IAmount := common.FloatToBigInt(amount, int64(asset.Decimals))
```
secmask, line 387: why caption var name?

exchange/binance.go
```diff
@@ -354,8 +370,168 @@ func (bn *Binance) GetTradeHistory(fromTime, toTime uint64) (common.ExchangeTrad
...
+			tx.Hash().Hex(),
+			bn.ID().String(),
+			asset.ID,
+			common.MiningStatusFailed,
+			common.ExchangeStatusFailed,
+			amount,
+			common.GetTimestamp(),
+		)
+		// update tx to activity, set blockchain status to pending
+		return bn.storage.StoreIntermediateDeposit(id, data, false)
```
secmask, line 451: how do we use pending state, seem it does not use anywhere.

halink0803:
I think the PR is good to go.

secmask (approved):

PR #1341, opened 2021-05-28
update log when tx status is not consistent with exchange status

halink0803:


exchange/binance.go
```diff
@@ -623,6 +623,13 @@ func (bn *Binance) WithdrawStatus(id string, assetID rtypes.AssetID, amount floa
 				if err != nil {
 					bn.l.Errorw("get withdraw fee failed: %w", err)
 				} // in case error we just make it pass and check err manual later, so it wont stuck on withdraw
+				txStatus, _, err := bn.binBlockchain.TxStatus(ethereum.HexToHash(withdraw.TxID))
```
secmask, line 626: it's better to check this some where in saveAuth data, where we already had blockchain + exchange status, so we dont have to check blockchain status again

secmask (approved):

PR #1353, opened 2021-07-13
add 0x setting

secmask:


secmask:
@ductm54 I added `enabled` field.

reservesetting/storage/postgres/assets.go
```diff
@@ -1041,6 +1074,15 @@ func (s *Storage) updateAsset(tx *sqlx.Tx, id rtypes.AssetID, uo storage.UpdateA
 		SanityRateProvider:    sanityRateProvider,
 		SanityRatePath:        sanityRatePath,
 	}
+	if uo.ZeroX != nil {
+		arg.updateZeroXPref = updateZeroXPref{
```
NNagato, line 1078: please check and add this param to `updateMsgs`, cuz, if `updateMsgs` is empty it's, will return an error in line 1194

NNagato (changes requested):

reservesetting/storage/postgres/assets.go
secmask: nice catch, I fixed it.

NNagato (approved):
LGTM

PR #1358, opened 2021-08-16
Convert eth weth

secmask:


secmask:
changes made in this MR: 
- reserve contract no longer use, however we still keep its name but actually will be external wallet(multisig) that use to fetch reserve balances, will rename when frontend ready for changes.
- withdraw from binance will take an extra step that convert ETH -> WETH
- deposit now use separate contract called operation-contract.

halink0803 (approved):
LGTM

PR #1377, opened 2021-09-27
fix issue where activity turn to final when withdraw tx mined, this is a bug after update wrapETH tx

secmask:
fix issue where activity turn to final when withdraw tx mined, this is a bug after update wrapETH tx because we IsPending() set to false when MiningStatus set to mined, fix to check with ExchangeStatus aswell


halink0803 (approved):
Lgtm

